PR #18, opened 2015-03-29
Fixed deprecation error for VC++ in C generator

Shimmen:
Visal C++ complier has deprecated sscanf() and replaced it with sscanf_s(). Using the former will cause a compilation error.


Dav1dde:
I am sorry that this simple change took so long .. I did exactly what you did in your PR, couldn't merge it since I rebased the whole projects.

Thanks!

PR #20, opened 2015-05-15
CmakeLists.txt that generates glad

hrehfeld:
I added a sensible cmakelists.txt that calls python to generate glad and creates the lib.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,33 @@
+cmake_minimum_required(VERSION 3.0)
+project(GLAD)
+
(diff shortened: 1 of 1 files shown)


Dav1dde:
Thanks for the pull request, would be awesome if you could change the commandline and squish it down to one commit.

hrehfeld:
from the FindPythonInterp.cmake I think Python_ADDITIONAL_VERSIONS just suggests additional versions (and i'm using it mixed with python 3 here)

hrehfeld:
done.

PR #21, opened 2015-05-15
add: CmakeLists.txt that generates glad

hrehfeld:


Dav1dde:
Thanks! Just for future reference, followup of #20.

PR #22, opened 2015-06-29
Get OpenGL extensions strings only once.

jrfonseca:
Invoke `glGetString(GL_EXTENSIONS)` only once (instead of
GL_NUM_EXTENSIONS), and `glGetStringi(GL_EXTENSIONS)` only
GL_NUM_EXTENSIONS (instead of roughly
GL_NUM_EXTENSIONS*GL_NUM_EXTENSIONS).

This is particularly handy when obtaining traces of applications that
use GLAD, which otherwise get over 10K of redundant calls nowadays
(and probably much more in the future).

### DIFF
--- a/glad/lang/c/generator.py
+++ b/glad/lang/c/generator.py
@@ -79,6 +79,7 @@ class CGenerator(Generator):
 
             f.write('static void find_extensions{}(void) {{\n'.format(api.upper()))
             if self.spec.NAME in ('gl', 'glx', 'wgl'):
+                f.write('\tget_exts();\n')
                 for ext in extensions[api]:
                     f.write('\tGLAD_{0} = has_ext("{0}");\n'.format(ext.name))
             f.write('}\n\n')
@@ -216,7 +217,7 @@ class CGenerator(Generator):
                 f.write('GLAPI int gladLoad{}Loader(GLADloadproc);\n\n'.format(api.upper()))
 
     def write_code_head(self, f):
-        f.write('#include <stdio.h>\n#include <string.h>\n#include {}\n'.format(self.h_include))
+        f.write('#include <stdio.h>\n#include <stdlib.h>\n#include <string.h>\n#include {}\n'.format(self.h_include))
 
     def write_extern(self, fobj):
         fobj.write('#ifdef __cplusplus\nextern "C" {\n#endif\n')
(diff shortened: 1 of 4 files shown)


Dav1dde:
Thanks! That's a good change, unfortunatly it breaks under certain conditions:

```
build/src/glad_egl.c: In function ‘find_extensionsEGL’:
build/src/glad_egl.c:225:2: error: implicit declaration of function ‘get_exts’ [-Werror=implicit-function-declaration]
```

Using:
```
python -m glad --generator=c --spec=egl --out-path=build
gcc -Ibuild/include build/src/glad_egl.c -o build/tmp.o -Wall -Werror -ansi -c
```

You can run the `utility/compiletest.sh` for some basic tests (to not download the XML files every time use `utility/compiletest.sh no-download`).

jrfonseca:
I've pushed an updated commit.

I decided to introduce empty `get_exts` stubs for EGL/WGL/GLX, but I believe sooner or later it will be useful to implement those stubs and generalize the extension string caching for those windows-specific extensions strings too.

Dav1dde:
I am not sure if saving the extensions globally without freeing the array is specifically nice, but it's definitly easier. Maybe if I have some more time I can come up with something. I guess having a few bytes of memory allocated isn't that bad (would valgrind even complain?).

I definitly agree, filling in the stubs would be good next step, might look into that if I have some free time. I am merging this and opening an issue for the stubs (feel free to fix this for me ;)).

Just some fun 'statistics':
Pre commit: 136750 glGetStringi calls
After commit: 337

That's a really good change, thanks!

jrfonseca:
> would valgrind even complain?

No. I've doubled checked here, and valgrind does not complain, as the link to memory is still there on the global, and this is common practice. Only if one passes `--show-reachable=yes` option will it report, but when one passes that option, valgrind will report all sort of stuff (fe.g, from Xlib, Mesa, etc.)

It would be trivial to free this with a C++ destructor, but with standard C there's not much alternative, other than have a `gladUnload` entry-point. 

I'm using GLFW, so the EGL/GLX/WGL extensions caching there doesn't affect me directly yet.

Thanks for pushing.  And thanks for GLAD.  I was using GLEW for ApiTrace unit tests, but the lack of GLES support was frustrating me. GLAD is really simple to use and embedded.

Dav1dde:
I was thinking about something like:

```
load_extionsions() {
    get_exts();
    ... magic here ...
    free_exts();
}
```

Even better would be not using global state at all and passing the list around, I gotta think about it, but the current solution is definitly fine, especially if valgrind doesn't even complain.

I definitly want to avoid a gladUnload function of any kind (and making it optional is just stuipid).

Glad you like it (I love these puns) :).

PR #26, opened 2015-08-17
cmake: fix typo in GLAD_SOURCES

aaronmjacobs:
The path to the glad header is incorrect, causing the custom command to regenerate the source files on every build. This simply fixes the path, so that the command only generates the sources when they are missing.


Dav1dde:
Thank you!

PR #27, opened 2015-08-17
cmake: add support for options

aaronmjacobs:
This change allows source generation options to be configurable through CMake. For example, if an OpenGL 3.2 core profile was desired:
```
set(GLAD_PROFILE "core" CACHE INTERNAL "OpenGL profile")
set(GLAD_API "gl=3.2" CACHE INTERNAL "API type/version pairs, like \"gl=3.2,gles=\", no version means latest")
add_subdirectory("${PROJECT_SOURCE_DIR}/dependencies/glad")
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,33 +1,73 @@
 cmake_minimum_required(VERSION 3.0)
 project(GLAD)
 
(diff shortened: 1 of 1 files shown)


Dav1dde:
Thanks!

PR #35, opened 2015-10-13
Correct the placement of call-convention specifier

MiniAl:
No-one seems to have labelled this as a bug, but on Visual Studio code generated with the `c-debug` generator fails to compile. I've tracked this down to the substitution of `APIENTRY` to `__stdcall`.

This means that the calling convention is specified in the wrong place? Swapping the placement of the type-specifier, and the call convention fixes this. Function pointers also already have theirs specified in their `typedef`.

The `c-debug` generator results now correctly compile for me on Visual Studio 2013.


Dav1dde:
Thank you very much, I manually pushed it (only changed the name of the commit).

PR #37, opened 2015-10-17
c: Fix a directive of GLVersion for access in DLL.

bluekyu:
When C DLL version is used, GLVersion variable cannot be accessed because of omitting export. I changed "extern" to "GLAPI" and moved the statement to the next of API generation.


Dav1dde:
Looks good, thanks!
In a few days all the changes I made in the last days (including this) will be live on the webgenerator, after some testing!

PR #38, opened 2015-11-06
Fix running GLAD locally with pip installed.

trevex:
I use GLAD as ExternalProject_Add CMake dependency and ran into an issue in environments,
where pip is available but glad is not installed, because pkg_resources is available, but raises an exception.


Dav1dde:
Fixed in 961af21c224233cec27dad3da4a1b6ce359d0828 - Thank you very much!

PR #39, opened 2015-11-07
Source version from a common point

yerejm:
To build upon 961af21c224233cec27dad3da4a1b6ce359d0828: it would still be nice to have the version in the generated source. Would a [common version file](https://stackoverflow.com/questions/458550/standard-way-to-embed-version-into-python-package) that can be sourced by both glad and packaging be acceptable?


Dav1dde:
Thanks!

PR #47, opened 2016-04-01
Fix the handling of the --no-loader flag

johnnovak:
Without this fix `Loader: No` will always appear in the file comments of the generated files.


Dav1dde:
Good catch, thanks! (I manually pushed the fix)

johnnovak:
:+1:

PR #49, opened 2016-04-20
Improve the C generator to especially better support for the Windows platform

QUSpilPrgm:
Basicly I have made 3 changes to improve the usage of the C generator especially on Windows:
- `--local-files`: This command line option informs Glad to put all files into a single directory. Inside the MSVC environment it is uncommon to split files into seperate directories for headers and sources. Additionally it is no longer necessary to add the Glad header directory to the include paths which is much more convenient.
- `--omit-khrplatform`: This option prevents the often unnecessary file `khrplatform.h` from getting downloaded and included.
- I changed `LoadLibraryA` to `LoadLibraryW` since the usage of the old Ansi functions is generally not recommended anymore.

It would propably make sense to add some of these options to other generators but this is outside the scope of this pull request.

Note: I am not familiar with Python programming. If there is anything that could be improved, I would be happy to change it.

### DIFF
--- a/glad/lang/c/generator.py
+++ b/glad/lang/c/generator.py
@@ -17,22 +17,31 @@ class CGenerator(Generator):
         if not self.spec.NAME == 'gl':
             suffix = '_{}'.format(self.spec.NAME)
 
-        self.h_include = '<glad/glad{}.h>'.format(suffix)
-        self._f_c = open(make_path(self.path, 'src',
-                                   'glad{}.c'.format(suffix)), 'w')
-        self._f_h = open(make_path(self.path, 'include',
-                                   'glad', 'glad{}.h'.format(suffix)), 'w')
+        if self.local_files:
+            self.h_include = '"glad{}.h"'.format(suffix)
+            self._f_c = open(make_path(self.path,
+                                        'glad{}.c'.format(suffix)), 'w')
+            self._f_h = open(make_path(self.path, 
+                                        'glad{}.h'.format(suffix)), 'w')
+            khr = self.path
+        else:
+            self.h_include = '<glad/glad{}.h>'.format(suffix)
+            self._f_c = open(make_path(self.path, 'src',
+                                        'glad{}.c'.format(suffix)), 'w')
+            self._f_h = open(make_path(self.path, 'include', 'glad',
+                                        'glad{}.h'.format(suffix)), 'w')
+            khr = os.path.join(self.path, 'include', 'KHR')
 
-        khr_url = KHRPLATFORM
-        if os.path.exists('khrplatform.h'):
-            khr_url = 'file://' + os.path.abspath('khrplatform.h')
+        if not self.omit_khrplatform:
+            khr_url = KHRPLATFORM
+            if os.path.exists('khrplatform.h'):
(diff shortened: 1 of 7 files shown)


glad/lang/c/generator.py
```diff
@@ -5,6 +5,7 @@
 from glad.lang.common.util import makefiledir
 
 
+
```
Dav1dde: nitpick

glad/parse.py
```diff
@@ -74,8 +74,13 @@ def comment(self):
     @property
     def types(self):
         if self._types is None:
-            self._types = [Type(element) for element in
-                           self.root.find('types').iter('type')]
+            elements = self.root.find('types').iter('type')
```
Dav1dde: Modifying the parser is incorrect here. I don't want the parser behaviour to change.

glad/parse.py
```diff
@@ -143,14 +148,17 @@ def extensions(self):
...
+    def __init__(self, element, local_files):
         apientry = element.find('apientry')
         if apientry is not None:
             apientry.text = 'APIENTRY'
         self.raw = ''.join(element.itertext())
         self.api = element.get('api')
         self.name = element.get('name')
 
+        if local_files:
+            self.raw = self.raw.replace('<KHR/khrplatform.h>', '"khrplatform.h"')
```
Dav1dde: Same here, the parser behaviour gets modified.

QUSpilPrgm:
Thanks for your fast reply.
Your comments seem reasonable. I have updated the code accordingly.

Dav1dde:
Thanks, looks good!

I somehow forgot to send my reply I typed yesterday... Anyways can you rebase the commits into one and prefix it with `glad: `? You can easily rebase with `git rebase -i HEAD~7` then select fixup or squash for the last commits.

I will merge once I've tested it locally, hopefully by next week.

QUSpilPrgm:
I have rebased the commits into a single one but i am not sure about how to change the prefix. What do you mean exactly? I am not using Git too often.

Dav1dde:
`git rebase -i HEAD~2` Then you can replace the `pick` at your comment with `reword`, this allows you to change the commit message.

QUSpilPrgm:
The commit message is changed and prefixed with `glad:`.
I hope that is correct now.

Dav1dde:
Thank you very much! I will add these to the web interface after rewriting the glad a little, I don't like the architecture of glad at all, especially the configuration handling (passing commandline options through to everything). 

If you're interested in what I changed: `local_files` and `omit_khrplatform` are no instance variables and I added the options to the header section.

stfx:
Hey David, friendly reminder that those options are still missing in the web interface

stfx:
> I will add these to the web interface after rewriting the glad a little

@Dav1dde those options are still missing in the web interface

Dav1dde:
Yes and they won't be added. I am currently completely rewriting glad, so I am not adding these options to this version.

stfx:
@Dav1dde I've seen you recently added the `omit_khrplatform` option to the web interface. Did you think of exposing `local_files` now too?

Dav1dde:
No, but just did, might as well.

PR #50, opened 2016-04-30
cmake: allow cmake to use primary python

yerejm:
In CMakeLists.txt, this seems unnecessary:

`set(Python_ADDITIONAL_VERSIONS 2)`

Using pyenv where both python 2 and python 3 are present, but where python 3 is the default python in use, this line seems to make cmake favour python 2, resulting in failure to find python.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,6 @@ project(GLAD)
 
 set(GLAD_DIR "${PROJECT_SOURCE_DIR}")
 set(GLAD_OUT_DIR "${PROJECT_BINARY_DIR}")
-set(Python_ADDITIONAL_VERSIONS 2)
 find_package(PythonInterp REQUIRED)
 
 # Options


Dav1dde:
This was brought up in the original PR: https://github.com/Dav1dde/glad/pull/20#issuecomment-102455844

@hrehfeld do you remember why you added it and if it is safe to remove it?

As I understand it, it should tell CMake just to consider Python 2. Maybe `set(Python_ADDITIONAL_VERSIONS 2 3)` would be correct.

hrehfeld:
If GLAD works without it (i.e. python2 and 3), it should be safe to remove iirc.

yerejm:
If I understand correctly, the [current active python should be searched for first](https://github.com/Kitware/CMake/blob/master/Modules/FindPythonInterp.cmake#L93), but `Python_ADDITIONAL_VERSIONS` overrides this and those additional pythons are searched for first (in order) instead.

So adding 3 to the end of `Python_ADDITIONAL_VERSIONS` won't solve this issue. It will still find `python2` instead of the active python.

`Python_ADDITIONAL_VERSIONS` appears to only be necessary when the active python is *not* the python that GLAD should use.

Dav1dde:
Ok, I couldn't find anything that says it is required. And it does work without on my system, I will merge this and test on a few different OSes, but it should be fine, since glad works with basically any python version.

Issue #44, opened 2016-02-25
Allow strict compiler flags (-Wall -Wextra -Werror -pedantic)


anderejd:
I will begin by saying thank you for this project! :)

Using -pedantic with GCC (tested with 5.3.0) will result in a lot of the following warnings (or errors for us using -Werror):
**ISO C forbids conversion of object pointer to function pointer type**

I solved it by modifying my generated glad.h:

```
typedef void (*GLADvoidfn)(void);
typedef GLADvoidfn (*GLADloadproc)(const char *name);

```

This will avoid the casts from pointer-to-void to pointer-to-function inside glad.

In my calling code I also needed to add this since I use SDL which also uses pointers-to-void for function pointers:

```
static GLADvoidfn
getprocaddress_wrapper(const char *name)
{
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-pedantic"
    return (GLADvoidfn) SDL_GL_GetProcAddress(name);
#pragma GCC diagnostic pop
}
```

I have not tried this on other compilers yet.
Would you like a pull request?

Dav1dde:
Hey, thanks for the report. I could reproduce the problem and looked further into the issue. It turns out ISO C does not guarantee casts from `void*` to function pointers to succeed, since the standard assumes `void*` is a different type (data) than a function pointer and these can have different representations. [Steve Summit explains that better than me, in his Stackoverflow answer](http://stackoverflow.com/a/31527867/969534).

There are workarounds to trick the compiler (see SO answer), or your solution which only shifts the problem to the loader function. 

The only real solution is to compile the relevant files (glad) without `-pedantic`, simply because casting from `void*` to a function pointer is not ISO C semantics.

rajder:
Hi!

Yes, I searched for the standard info too and found the same thing, void pointers are pointers to object and casting to function pointer is implementation defined. (however Posix does define the cast, but that is still not in the ISO standard)

I'm not sure that I agree with your conclusion, the implementation defined behavior is still there even if one would turn off the warnings in the compiler, so I would not call that a solution.

My thinking is that moving the necessary cast (at least working with SDL) outside glad would at least keep glad compliant to the standard and leave the decision about lowering warning levels vs tricking the compiler to be silent up to the user.

Best Regards /Anders

Dav1dde:
I think you are right, changing the return type of the loader might just work, if you want you can make a pull request.

[...]

PR #56, opened 2016-07-17
Support for strict c compiler flags.

anderejd:
Tested on:
- GCC, Linux, -Wall -Wextra -pedantic -Werror
- Clang, OS X, -Wall -Wextra -pedantic -Werror
- MSVC 2015, Windows 10, /W4 /WX

### DIFF
--- a/glad/lang/c/loader/__init__.py
+++ b/glad/lang/c/loader/__init__.py
@@ -1,9 +1,10 @@
 
 LOAD_OPENGL_DLL = '''
-%(pre)s void* %(proc)s(const char *namez);
+%(pre)s GLADvoidfn %(proc)s(const char *namez);
 
 #ifdef _WIN32
 #include <windows.h>
+#include <string.h>
 static HMODULE libGL;
 
 typedef void* (APIENTRYP PFNWGLGETPROCADDRESSPROC_PRIVATE)(const char*);
@@ -29,11 +30,18 @@ void %(terminate)s(void) {
     }
 }
 #else
+#include <string.h>
 #include <dlfcn.h>
 static void* libGL;
 
 #ifndef __APPLE__
-typedef void* (APIENTRYP PFNGLXGETPROCADDRESSPROC_PRIVATE)(const char*);
+/*
+ * The definition of glXGetProcAddressARB is found here:
+ * https://www.opengl.org/registry/ABI/
+ * https://www.opengl.org/registry/specs/ARB/get_proc_address.txt
+ * void (*glXGetProcAddressARB(const GLubyte *procName))();
+ */
+typedef void (*(*PFNGLXGETPROCADDRESSPROC_PRIVATE)(const GLubyte *procName))();
 PFNGLXGETPROCADDRESSPROC_PRIVATE gladGetProcAddressPtr;
 #endif
 
@@ -48,6 +56,7 @@ int %(init)s(void) {
     };
 #else
     static const char *NAMES[] = {"libGL.so.1", "libGL.so"};
+    void *pv = NULL;
 #endif
 
     unsigned int index = 0;
@@ -58,8 +67,8 @@ int %(init)s(void) {
 #ifdef __APPLE__
             return 1;
 #else
-            gladGetProcAddressPtr = (PFNGLXGETPROCADDRESSPROC_PRIVATE)dlsym(libGL,
-                "glXGetProcAddressARB");
+            pv = dlsym(libGL, "glXGetProcAddressARB");
+            memcpy((void *)&gladGetProcAddressPtr, (void *)&pv, sizeof(pv));
             return gladGetProcAddressPtr != NULL;
 #endif
         }
@@ -77,21 +86,32 @@ void %(terminate)s() {
 }
 #endif
 
+/*
+ * TODO: Split function into a windows version and one nix version?
+ *       Would probably make it more readable/maintainable?
+ */
 %(pre)s
-void* %(proc)s(const char *namez) {
-    void* result = NULL;
+GLADvoidfn %(proc)s(const char *namez) {
+    GLADvoidfn result = NULL;
+    void *pv = NULL;
     if(libGL == NULL) return NULL;
 
 #ifndef __APPLE__
     if(gladGetProcAddressPtr != NULL) {
-        result = gladGetProcAddressPtr(namez);
+#ifdef _WIN32
+        pv = gladGetProcAddressPtr(namez);
+        memcpy((void *)&result, (void *)&pv, sizeof(pv));
+#else
+        result = gladGetProcAddressPtr((const GLubyte *)namez);
+#endif
     }
 #endif
     if(result == NULL) {
 #ifdef _WIN32
-        result = (void*)GetProcAddress(libGL, namez);
+        result = (GLADvoidfn)GetProcAddress(libGL, namez);
 #else
-        result = dlsym(libGL, namez);
+        pv = dlsym(libGL, namez);
+        memcpy((void *)&result, (void *)&pv, sizeof(pv));
 #endif
     }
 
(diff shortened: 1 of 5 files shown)


Dav1dde:
Thanks for that, I will have a look at it when I am near a PC again (~in a week).

rajder:
Here is an example of how to use SDLs loader with the strict_flags branch:

```
/**
 * The purpose of this function is to help enable strict compiler flags,
 * without warnings.
 *
 * Exploit that pointer-to-pointer-to-function is a pointer-to-object and can
 * be cast to pointer-to-pointer-to-void, then dereferenced and written to. We
 * are sneaking in the cast from pointer-to-function to pointer-to-void by one
 * extra step of indirection. The cast is still implementation defined
 * behavior, but must be done since SDL_GL_GetProcAddress returns a
 * pointer-to-void.  The alternative is to use compiler specific pragmas to
 * ignore the warning.
 */
GLADvoidfn
wrap_sdl_gl_getprocaddress (const char *name)
{
	GLADvoidfn f;
	*(void **) (&f) = SDL_GL_GetProcAddress (name);
	return f;
}

...

	if (!gladLoadGLLoader (wrap_sdl_gl_getprocaddress))
	{
		printf ("glad failed to load!")
...

```

The built-in glad loader should work without changes in user code.

rajder:
...Maybe it would be better to put the three cast in the example wrapper function inside gladLoadGLLoader instead? That would keep the API to glad unchanged.

Dav1dde:
Thanks for your work again, I will not merge this PR but I will definitely look into it and get it into the `glad2` branch if it makes sense and works.

rajder:
Ok cool!
Then I will probably keep making small changes in this branch for my own use, like keeping the signature identical for gladLoadGLLoader and possibly adding a gladLoadGLLoaderStrict with the new signature.

QUSpilPrgm:
I am pretty sure casting a pointer to a different type and writting to it is actually undefined behaviour in C and C++ due to strict aliasing.
This seems to make things worse but maybe I am missing something.

rajder:
@QUSpilPrgm Thank you for the input. I think you are right, but the cast must be done at some point since in the SDL and linux case we get the gl function pointers as pointer-to-void and must be cast to the correct pointer-to-function type before calling the function through the pointer. One solution is to lower the compiler warnings and do the cast directly from pointer-to-void to pointer-to-function, but with strict compiler flags we need to use compiler specific pragmas or use one extra level of indirection as explained in the doc comment in the example: https://github.com/Dav1dde/glad/pull/56#issuecomment-233281738

I think I have done my homework on this but I would also appreciate more feedback.

rajder:
@QUSpilPrgm The more I read up on strict aliasing the more this seems to be a bad idea.

I will experiment with enabled strict aliasing and look at the generated code. I suspect that I will abandon this attempt and simply accept lowering warning levels and do the cast. Thanks for the feedback :)

QUSpilPrgm:
Yes you are right. Some cast must be done at some point, but I don't think it has to violate strict aliasing. Here is my understanding of the situation:

Strict aliasing means that a type must not be accessible by a pointer of a different type that is not either `void*` or `[signed|unsigned] char*`. In this case you are casting `PFNGLXGETPROCADDRESSPROC_PRIVATE*` to `void**`. `void**` is a different type and not one of the exceptions. Note that a pointer to anything (`void*`) is different to a pointer to a pointer (`void**``). 

Previously there was a cast from `void*` to `void* (*)(const char*)`(`PFNGLXGETPROCADDRESSPROC_PRIVATE`) which seems ok, since `void*` can be a pointer to anything. What it actually points to is outside the scope of C and Glad. You could imagine that somewhere there is actually a function of the type `void* ()(const char*)`. A pointer to it was casted to `void*` somewhere just to return it from `dlsym`. Inside Glad it is than casted back to `void* (*)(const char*)`. Thats all fine because the function was never reachable by a pointer which is not one of the exceptions. This seems similar to casting the result of `malloc` to anything you want for the first time.

rajder:
The pull request is an attempt to avoid "iso c forbids conversion of object pointer to function pointer", which is raised when casting from pointer-to-void to pointer-to-function. That warning is avoided but I think you might be right about breaking strict aliasing, but I'm not sure yet.

QUSpilPrgm:
Oops, sorry, I didn't realize that function pointers are generally incompatible with `void*`. I guess that makes sense for Harvard architectures where the programm memory is bigger than the RAM. That means that the `dlsym` usage will always be a bit iffy. That seems to be a bug in POSIX.

But maybe there are still better ways. How about using `memcpy` or a `union` to avoid the strict aliasing issue?
Like this:
` void* getProcAddressPtrVoidStar = dlsym(libGL, "glXGetProcAddressARB");
memcpy(&gladGetProcAddressPtr, &getProcAddressPtrVoidStar, sizeof(void*));`
Or this in C: (The union violates C++'s strict aliasing)
`union { void* void_star; PFNGLXGETPROCADDRESSPROC_PRIVATE func_ptr; } getProcAddressPtrTemp;`
`getProcAddressPtrTemp.void_star = dlsym(libGL, "glXGetProcAddressARB");`
`gladGetProcAddressPtr = getProcAddressPtrTemp.func_ptr;`

rajder:
nice idea :) but memcpy would result in the same suspected strict aliasing breakage, right?
The union trick would do for my current project but I suspect many users want C++ support.

The question is, does this violate strict aliasing?:
```
typedef void (*GLADvoidfn)(void);
...
    GLADvoidfn f;
    *(void **)(&f) = SDL_GL_GetProcAddress(name);
    return f;
...
```
I'm guessing that f could be returned as garbage if the compiler were to decide that f will not be affected by the assignment since the pointer to f has a different type than the pointer being dereferenced and written to?

rajder:
Nice, I will read those articles later today. Thanks!

rajder:
Nice read. It seems like memcpy or using char pointers and copy the bytes manually is the way to go to preserve C++ compatibility.

memcpy it is, I will update the branch later this week. Thanks again!

Dav1dde:
I will have to dig into this topic more at some point, but are there *any* disadvantages of the memcpy approach?

If not this may as well be the default implementation.

rajder:
We are still breaking ISO C by converting a object pointer to a function pointer, but that needs to happen since dlsym is what it is.

string.h need to be included, if that is not ok we could write a loop to do the copying of the pointer bytes using char pointers.

Not realated to memcpy, but a reault of this PR:
The API to glad, the signature of gladLoadGLLoader is not as comfortable as before. Now the user needs to be aware of how to wrap up SDL_GL_GetProcAddress and similar loader functions. New example:
```
static GLADvoidfn
sdl_gl_getprocaddress_wrapper (const char *name)
{
	GLADvoidfn f = NULL;
	void *pv = SDL_GL_GetProcAddress (name);
	memcpy (&f, &pv, sizeof(pv));
	return f;
}
```

Dav1dde:
Including string.h is fine.

About the API, it should be possible to do the `memcpy` within glad, but I don't know if that's desirable.

glad/lang/c/loader/__init__.py
```diff
@@ -78,20 +81,23 @@
...
+GLADvoidfn %(proc)s(const char *namez) {
+    GLADvoidfn result = NULL;
+    void *pv = NULL;
     if(libGL == NULL) return NULL;
 
 #ifndef __APPLE__
     if(gladGetProcAddressPtr != NULL) {
-        result = gladGetProcAddressPtr(namez);
+        pv = gladGetProcAddressPtr(namez);
+        memcpy((void *)&result, (void *)&pv, sizeof(pv));
```
rajder: When gladGetProcAddressPtr is glXGetProcAddressARB we can can to GLADvoidfn without memcpy and the extra void pointer. TODO: fix

rajder:
I looked into moving the wrapper inside glad but it did not seem trivial at a glance and I too am not sure if it was a good idea.

rajder:
Oops, forgot to test this last commit on windows. Fixing now.

rajder:
Rebased on master and squashed branch to one commit.

Dav1dde:
I finally implemented this for glad2, see #44 for an explanation.

PR #57, opened 2016-07-23
d: change deprecated imports to non-deprecated ones for windows api

profan:
Old import was still there in two places, the deprecation messages were bothering me a bit so I decided to make a PR. It's used in other places in the code already too, so it *should* be fine.


Dav1dde:
Thanks!

PR #59, opened 2016-08-03
Sorted extensions to make diff of generated code easier to read.

rajder:
The purpose is to minimize the length of the diff rows and the number of diff rows in the generated code.

### DIFF
--- a/glad/lang/common/generator.py
+++ b/glad/lang/common/generator.py
@@ -99,9 +99,9 @@ class Generator(object):
                 )
 
         if self.extension_names is None:
-           self. extension_names = list(chain.from_iterable(self.spec.extensions[a]
+           self.extension_names = list(chain.from_iterable(self.spec.extensions[a]
                                                             for a in self.api))
-
+	self.extension_names.sort()
         e = list(chain.from_iterable(self.spec.extensions[a] for a in self.api))
         for ext in self.extension_names:
             if ext not in e:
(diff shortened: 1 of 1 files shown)


Dav1dde:
Thanks for the pull, I took the diff and slightly changed it (newline in commandline doesn't work for copy pasting), I also found a bug where extensions where listed more than once.

rajder:
Yeah that's why I added the backslash for line continuation in bash. But on windows that needs to be ^ so not perfect :)

glad/lang/common/generator.py
```diff
@@ -182,7 +182,7 @@ def commandline(self):
         generator = '--generator="{}"'.format(self.NAME)
         specification = '--spec="{}"'.format(self.spec.NAME)
         loader = '' if self.has_loader else '--no-loader'
-        extensions = '--extensions="{}"'.format(','.join(self.extension_names))
+        extensions = '--extensions="\\\n            {}"'.format(',\\\n            '.join(self.extension_names))
```
rajder: backslash for line continuation in bash.

Dav1dde:
Yes, but as you stated doesn't work on Windows and in my opinion there is no additional value in formatting it, since there is already the formatted (and copy pasteable) list above.

rajder:
True, the command line blob can be ignored when viewing the diff.

PR #69, opened 2016-11-23
CMake: use target_include_directories(PUBLIC)

ranjak:
This causes GLAD_INCLUDE_DIRS to be automatically added to the include directories of the target that glad is being linked to.

E.g I only have to type `target_link_libraries(my_target glad)`, I don't need to add `target_include_directories(my_target PRIVATE ${GLAD_INCLUDE_DIRS})`.

The "old" way still works of course, this won't break anything AFAICT.


ranjak:
Updated with @rwols 's tweaks.

Dav1dde:
Thanks for your work, I hope I find time this weekend to merge all open pull requests.

Dav1dde:
Pushed in 337695bc80bdd7d49692b01f21246b6f5398bea2 - I might have broken it, can you check if it still works (Check the latest version: [master](https://github.com/Dav1dde/glad/blob/master/CMakeLists.txt))? Thanks!

rwols:
I just upgraded the glad submodule in my project and it compiles fine here!

Dav1dde:
Great, thank you!

PR #76, opened 2017-01-30
gl: remove unnecessary cast

dcousens:
Removes warning:

> src/glad.c:177:23: error: cast from type ‘const char**’ to type ‘char**’ casts away qualifiers [-Werror=cast-qual]

AFAIK,  I don't think this cast is necessary anyway?


Dav1dde:
Thanks! No idea why that cast was there.
Pushed in https://github.com/Dav1dde/glad/commit/7fc26161a2414e9408f3187797572d3cf20b46c7
(not a fan of small merge commits)

PR #77, opened 2017-01-30
gl: remove implicit casts, resolves warnings

dcousens:
This may not be the best way to go about this,  but it was just a resolution for a stricter code base which uses `-Wsign-conversion`.


Dav1dde:
Thanks! Pushed in https://github.com/Dav1dde/glad/commit/b1e9e2fa67ad2e0372608a2bea6787970c9cf260

PR #85, opened 2017-05-15
Too much spaces in D example.

vaspaese:
Remove spaces from D example.


Dav1dde:
Thanks for the PR!

The spaces are on purpose, they indicate at least one indentation level, since there are not toplevel statements in D. Indentation is the same in the C/C++ example.

PR #88, opened 2017-06-09
Fix the const warning for the compiler

YuhsiangLu:
Fix the compiler wrong when try to free the const type.


Dav1dde:
Thank you!

Pushed in: 3ee4f2a330059522eceafcafd7ee86061f5bef0c

Issue #96, opened 2017-07-17
EGL generator doesn't provide runtime flags for extensions


mefesto:
I recently used GLAD for EGL (1.4) and required a couple extensions:
 - EGL_KHR_create_context
 - EGL_KHR_debug

Those particular extensions aren't very important but just used as an example. I was expecting things to work like GLAD for OpenGL where I can check the runtime extension flags to see if the device the app is running on supports them and augment behavior accordingly. So I expected the be able to do this:

```cpp
    if (GLAD_EGL_KHR_debug) { /* setup debug stuff */ }
```

But it seems that the glad_egl.h doesn't declare an extension variable for this purpose. Is this a bug or, for EGL, are users expected to do something else?

Thanks

Dav1dde:
Hey, yeah this currently doesn't work, I have the generation of these flags for EGL turned off, I don't remember why I did it - Seems like an oversight (-> bug).

If it is an easy fix, I might add it, otherwise if you feeling lucky you can try the glad2 branch (which has these flags).

Dav1dde:
I am gonna close this as wontfix now, this will be supported in glad2.

PR #98, opened 2017-07-18
Added EGL runtime extension flags support

mefesto:
I had some spare time today to look at issue #96 a bit. I'm not terribly familiar with the project so I'm not sure if this was the best approach but it does seem to work -- at least in an EGL project of mine.

The tricky thing was that in EGL you need an EGLDisplay if you want to get the list of extensions. So I had to augment the generated loaders to require the user to pass the display. Below is an example of its use:

```cpp
EGLDisplay display = eglGetDisplay(EGL_DEFAULT_DISPLAY);
eglInitialize(display, 0, 0);

if (!gladLoadEGL(display)) {
  fprintf(stderr, "Unable to load EGL via GLAD!\n");
  return;
}

printf("EGL_KHR_get_all_proc_addresses? %d\n", GLAD_EGL_KHR_get_all_proc_addresses);
printf("EGL_KHR_swap_buffers_with_damage? %d\n", GLAD_EGL_KHR_swap_buffers_with_damage);
printf("EGL_KHR_gl_colorspace (build support)? %d\n", EGL_KHR_gl_colorspace);
printf("EGL_KHR_gl_colorspace (runtime support)? %d\n", GLAD_EGL_KHR_gl_colorspace);
```

And this is the command I used to generate the stubs:

    $ python3 -m glad --out-path /tmp/glad --api 'egl=1.4' --generator c --spec egl

Thanks

### DIFF
--- a/glad/lang/c/generator.py
+++ b/glad/lang/c/generator.py
@@ -69,7 +69,7 @@ class CGenerator(Generator):
             for feature in features[api]:
                 f.write('static void load_{}(GLADloadproc load) {{\n'
                         .format(feature.name))
-                if self.spec.NAME in ('gl', 'glx', 'wgl'):
+                if self.spec.NAME in ('gl', 'glx', 'wgl', 'egl'):
                     f.write('\tif(!GLAD_{}) return;\n'.format(feature.name))
                 for func in feature.functions:
                     f.write('\tglad_{0} = (PFN{1}PROC)load("{0}");\n'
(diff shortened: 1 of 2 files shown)


Dav1dde:
I think we discussed this on IRC already. Unfortunately this requires changing the loader API and I don't want any breaking changes in this glad version (that's what glad2 is for).

The issue will definitely be addressed in glad2.

PR #104, opened 2017-08-07
Fix race condition in CMakeLists.txt

rwols:
The "glad" target depends on the generated source files. It may happen happen
that these are not yet generated in a multithreaded build, so we make the
dependency explicit here. There's a new target called "glad-generate-files"
that requires the files to exist. The "glad" target depends on the new target.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,7 +54,7 @@ if(GLAD_NO_LOADER)
 endif()
 
 add_custom_command(
-  OUTPUT ${GLAD_SOURCES} 
+  OUTPUT ${GLAD_SOURCES}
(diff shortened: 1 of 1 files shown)


rwols:
I'd like to point out that this is a real problem, my other PRs (https://github.com/Dav1dde/glad/pull/105, https://github.com/Dav1dde/glad/pull/106) are just facelifts.

Dav1dde:
Thanks you very much, I will review the changes probably on friday!

Dav1dde:
Thanks!

Pushed in https://github.com/Dav1dde/glad/commit/3d3531a5cde975c47682d061eb487a49de7c23f0

PR #105, opened 2017-08-07
Remove the static requirement for the glad library

rwols:
You can build static libs by specifying `-DBUILD_SHARED_LIBS=OFF` for the
cmake invocation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,7 +54,7 @@ if(GLAD_NO_LOADER)
 endif()
 
 add_custom_command(
-  OUTPUT ${GLAD_SOURCES} 
+  OUTPUT ${GLAD_SOURCES}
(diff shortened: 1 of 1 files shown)


Dav1dde:
Thanks!

Pushed in https://github.com/Dav1dde/glad/commit/71ae3416bb69e5a069be61ec9341c6e607c741e5

rwols:
OK well, I'm getting a lot of undefined symbols now when I try to build glad as a dynamic library. Perhaps there's something going awry in the definition of `GL_API`? You could revert commit 71ae341 in the meantime.

If it turns out this is going to take a lot of work to fix, maybe just let it stay a static library because it's tiny anyway.

rwols:
I managed to fix it: https://github.com/Dav1dde/glad/pull/111

PR #106, opened 2017-08-07
Improve project() and set linker language to C

rwols:
Also, only export the glad variables when we are not the root of the project.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@ if(NOT CMAKE_VERSION VERSION_LESS 3.1)
   cmake_policy(SET CMP0054 NEW)
 endif()
 
-project(GLAD)
+project(GLAD VERSION 0.1.14 LANGUAGES C)
 
 set(GLAD_DIR "${CMAKE_CURRENT_SOURCE_DIR}")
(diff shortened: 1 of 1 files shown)


Dav1dde:
Is this a breaking change? 

Does ` +if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)` this automatically set the EXPORT variable under a certain condition, if so, why?

rwols:
> Does +if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR) this automatically set the EXPORT variable under a certain condition

Yes. If the `glad` source dir is the "root" of the project, it will *not* export the variables to the parent project (because there is none).

If the `glad` source dir is *not* the "root" of the project, for example, it was added by a call to `add_subdirectory(glad)` and glad is a git-submodule (in my case), then it will export the variables to the parent project.

The `CMAKE_CURRENT_SOURCE_DIR` is the directory of the current CMakeLists.txt file. The `CMAKE_SOURCE_DIR` is the directory of the root CMakeLists.txt file.

> Is this a breaking change?

I'm fairly certain it isn't, but I'll test things out just in case.

Dav1dde:
Thank you!

PR #120, opened 2017-11-06
Extension checks make copy of glGetStringi values

teburt:
#### Description

The GL C extension checking code currently assumes that the pointer returned from `glGetStringi` will be unique for the given arguments. This is not the case on all devices. It is possible the value of the pointer returned from `glGetStringi` will be the same regardless of arguments, even though the string it points to is correct after each call and until the next call is made.

In order to maintain the correct list of the extension strings the implementation supports, a copy of the string needs to be made after each call to `glGetStringi`

Note: **This change assumes extension strings are less than 256 characters**

String functions like `strcpy` and `strlen` are avoided due to portability/security concerns.

#### Known Affected platforms:

| Device | Version | GraphicsVendor | GraphicsRenderer | GraphicsVersion |
|---------|---------|-------------------|---------------------|-------------------|
| Samsung Galaxy S4 | 4.4.4; API 19 | Qualcomm | Adreno 320 | OpenGL ES 3.0 V@45.0
| Moto X | 5.1; API 22 | Qualcomm | Adreno 320 | OpenGL ES 3.0 V@84.0


Dav1dde:
Wow thanks for figuring this out. I will take a look, merge and release this tomorrow.

teburt:
No problem, glad to contribute back. 

So far I have only seen the behavior on some of the earlier drivers to support GLES 3. But yes, it does make glad think there are basically no extensions present.

Dav1dde:
> String functions like strcpy and strlen are avoided due to portability/security concerns.

I am already using `strlen` and other `string.h` functions, what is your concern with them?

teburt:
> I am already using strlen and other string.h functions, what is your concern with them?

Actually, none. This was more of a platform/build issue for us. In this case using `strcpy` caused Visual Studio to issue a warning when compiling for Windows Universal (where they suggest `strcpy_s`). I suppose it would be better to switch to the safe version (when it's available) than do it by hand like this.

teburt:
> Pushed in 067f2e4 - Can you look over it again?

That looks good, it's what I had before running into trouble with strcpy.

teburt:
Thanks for getting this in @Dav1dde - I will close out this PR since it's in master.

PR #133, opened 2018-02-25
cmake : Add library destination.

p-groarke:
Fixes a cmake error when building in shared mode.
```
install TARGETS given no LIBRARY DESTINATION for shared library target
```


Dav1dde:
Thanks! Couldn't reproduce the problem but looking through the documentation it seems like it should be there.

PR #140, opened 2018-03-22
Remove const from exts_i variable to avoid -Wcast-qual warnings

Hendricks266:
With `-Wcast-qual` enabled, glad.c gives the following warning:

```
source/glad/src/glad.c: In function 'void free_exts()':
source/glad/src/glad.c:202:38: warning: cast from type 'const char*' to type 'char*' casts away qualifiers [-Wcast-qual]
             free((char *)exts_i[index]);
                                      ^
```

Upon inspection, this variable is filled with an allocation on the heap, which suggests its type should not point to const data.

### DIFF
--- a/glad/lang/c/loader/gl.py
+++ b/glad/lang/c/loader/gl.py
@@ -29,7 +29,7 @@ static int max_loaded_minor;
 
 static const char *exts = NULL;
 static int num_exts_i = 0;
-static const char **exts_i = NULL;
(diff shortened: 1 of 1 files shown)


Dav1dde:
Thanks! 

I'll merge/look into it when I am back at home!

Dav1dde:
Thanks! Took me way longer that it should, but it's fixed now.

PR #143, opened 2018-04-01
Add pascal support

dpethes:
Pascal generator for GL, based on the Nim generator. It doesn't have any library loader, as I use SDL2 as a loader so I have no use for it. 
Works with Freepascal 3.0.x on win32/64, most likely on older versions and other OSes as well.

Edit: added extension loading

### DIFF
--- a/glad/__main__.py
+++ b/glad/__main__.py
@@ -3,7 +3,7 @@
 """
 Uses the official Khronos-XML specs to generate a
 GL/GLES/EGL/GLX/WGL Loader made for your needs. Glad currently supports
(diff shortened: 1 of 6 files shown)


Dav1dde:
That's awesome! Thanks for your work!

dpethes:
It should be pretty much on par with the Nim generator now. If you have any ideas for improvements, let me know :)

Dav1dde:
I manually rebased this into the master branch and released it in version 0.1.22.
Also added it to the online generator.

Thank you very much for the backend!

dpethes:
Thanks!

PR #154, opened 2018-06-26
Add `type` field to the `Enum` class

Rijenkii:
There is a `type` attribute on some enums, specifically: `GL_INVALID_INDEX` ("u"), `GL_TIMEOUT_IGNORED` and `GL_TIMEOUT_IGNORED_APPLE` ("ull"). Documentation avaliable [here](https://github.com/KhronosGroup/OpenGL-Registry/blob/master/xml/readme.pdf), on page 9, section 9.1


Dav1dde:
Thanks! I pushed it with https://github.com/Dav1dde/glad/commit/6773b2bd614d2b9b694a58418074f0092c9e8316.

PR #160, opened 2018-08-05
treat extensions with empty string as None to fix build using CMake

firodj:
I am trying build using CMake and it should be create with all extensions when extensions is empty.

### DIFF
--- a/glad/__main__.py
+++ b/glad/__main__.py
@@ -132,6 +132,8 @@ def main():
         ns.generator, spec.NAME.lower()
     )
 
+    extensions = ns.extensions if ns.extensions else None
+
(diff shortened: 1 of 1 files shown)


madebr:
How about assuming adding 2 mutually exclusive options: `--all-extensions` and `--extensions`?
And make `--all-extensions` the default.
Also:
- The command line in the header comments should reflect the actual command line used.
- The header can have an extensions section with a (sorted) list of all actually included extensions

Dav1dde:
Makes sense, this is probably the saner and backwards compatible approach.

Dav1dde:
Fixed with #230 / #232

Issue #109, opened 2017-08-11
glad2 Documentation


Dav1dde:
### C
- [x] Examples
- [ ] Quick Start
- [x] Options

### Rust
- [ ] Examples
- [ ] Quick Start
- [ ] Options

Related issues:

* #161

BlGene:
One question to glad2, am I doing this right? ```gladLoadGLSimple(GLADsimpleloadproc())```

Dav1dde:
Still deciding on the API, currently it's:

```
typedef void* (* GLADloadproc)(const char *name, void* userptr);
typedef void* (* GLADsimpleloadproc)(const char *name);

GLAPI int gladLoadGL(GLADloadproc load, void* userptr);
GLAPI int gladLoadGLSimple(GLADsimpleloadproc load);
GLAPI int gladLoadGLInternalLoader();
```

Maybe naming the functions according to what arguments they take is the better solution. My only set goal is to get rid of the `gladLoadGL()` which uses the internal loader, I want the default usage to be without loader (basically every windowing framework has a `GetProcAddress` implementation).

BlGene:
...So if I call gladLoadGLInternalLoader I don’t have to call gladLoadGLSimple?

Dav1dde:
Correct, you only need to call one of those three (calling these functions in total more than once may result in weird results).

BlGene:
ProTip: When calling the glad executable add `--loader`

The loader function names have also changed:
`gladLoadEGLInternalLoader` -> `gladLoaderLoadEGL`
`gladLoadGLInternalLoader` -> `gladLoaderLoadGL`

BlGene:
@Dav1dde 
I just spent a while figuring out that `gladLoaderLoadGL()` was failing on certain machines because it didn't find the `libGL.so.1` file. It would be nice to have more informative error information from this function.

Dav1dde:
Thanks, I am still working on the documentation, the wiki should be accurate: https://github.com/Dav1dde/glad/wiki/C 

Interesting that libGL.so isn't available, what else is available?

I am not sure how I would do a better error handling. Return code 0 (false) is failure, return code > 0 is the loaded version, negative return codes would break the `if (!load())` "idiom".

BlGene:
@Dav1dde:
It looks like the problem wasn't with `dlopen`.

[...]

PR #161, opened 2018-08-05
Rust support for glad2

Dav1dde:
This adds a Rust generator for glad2.

- [x] EGL
- [X] GL
- [X] GLX
- [X] Vulkan
- [X] WGL

This includes aliasing support.

In comparison to C this contains no extended loader functions only a way to populate all loading function pointers, no checks for versions and extensions, depending on demand this may be added in the future.

Todo: 

- [ ] Documentation


Dav1dde:
Tracking documentation in #109 and merging.

PR #163, opened 2018-08-10
c: sort extensions and function pointers

### LINKED ISSUE #162: Running glad twice is not reproducible
Running the following commands:
```
glad --out-path glad1 --generator c --profile core --omit-khrplatform --spec gl --api gl=3.3
glad --out-path glad2 --generator c --profile core --omit-khrplatform --spec gl --api gl=3.3
```
will generate 2 glad opengl 3.3 core loaders in the directories `glad1` and `glad2`.

The two directory trees are different however:
- they contain a different timestamp; and
- the order in which some global variables are defined in `glad.c` is different.

I would suggest to:
- add no timestamp by default or add a `--reproducible`/`--notimestamp` command line option; and
- explicitly order all variables.

Using `pip 0.1.26` from pypi

madebr:
Fixes #162 (for c sources)

### DIFF
--- a/glad/lang/c/generator.py
+++ b/glad/lang/c/generator.py
@@ -231,6 +231,7 @@ class CGenerator(Generator):
             for feature in features:
                 f.write('int GLAD_{} = 0;\n'.format(feature.name))
 
+        write = sorted(list(write), key=lambda x: x.proto.name)
         for func in write:
             self.write_function(f, func)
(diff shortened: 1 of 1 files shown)


Dav1dde:
Thanks, pushed in 7e816198a872435a6281df51b0428bc4c0945a87

PR #168, opened 2018-08-14
fix for GLsizeiptr type

krux02:
according to [this document](https://www.khronos.org/opengl/wiki/OpenGL_Type), ``GLsizeiptr`` has to be the size of a pointer. So I fixed that type definition.


Dav1dde:
Thanks! Can you prefix the commit with `nim: `?

Also interesting that `int` is variable size in nim.

krux02:
``int`` is always the size of a pointer in Nim, same as ``uint``. I don't know how to prefix the commit comment at this point in time.

Dav1dde:
Thanks!

PR #178, opened 2018-10-08
Fix MSVC static analysis warning on realloc

PedDavid:
Using Microsoft (R) C/C++ Optimizing Compiler Version 19.15.26730 for x64
with the `/analyze` flag for static analysis, a warning is issued:

glad.c(777) : warning C6308: 'realloc' might return null pointer: assigning null pointer to 'exts_i', which is passed as an argument to 'realloc', will cause the original memory block to be leaked.

This change avoids the warning.

### DIFF
--- a/glad/lang/c/loader/gl.py
+++ b/glad/lang/c/loader/gl.py
@@ -43,7 +43,8 @@ static int get_exts(void) {
         num_exts_i = 0;
         glGetIntegerv(GL_NUM_EXTENSIONS, &num_exts_i);
         if (num_exts_i > 0) {
-            exts_i = (char **)realloc((void *)exts_i, (size_t)num_exts_i * (sizeof *exts_i));
+            char **tmp_exts_i = (char **)realloc((void *)exts_i, (size_t)num_exts_i * (sizeof *exts_i));
+            exts_i = tmp_exts_i;
         }
 
         if (exts_i == NULL) {


glad/lang/c/loader/gl.py
```diff
@@ -43,7 +43,8 @@
         num_exts_i = 0;
         glGetIntegerv(GL_NUM_EXTENSIONS, &num_exts_i);
         if (num_exts_i > 0) {
-            exts_i = (char **)realloc((void *)exts_i, (size_t)num_exts_i * (sizeof *exts_i));
+            char **tmp_exts_i = (char **)realloc((void *)exts_i, (size_t)num_exts_i * (sizeof *exts_i));
+            exts_i = tmp_exts_i;
```
Dav1dde: This is a workaround for the problem but doesn't seem to solve it, that if the reallocation fails the original memory is leaked. I think we would have to free the original memory then return with an error code.

Luflosi:
As far as I can tell, this commit doesn't actually fix the memory leak the compiler warned about. For that there would need to be a `free(exts_i);` above the `return 0;`. The code before the commit does exactly the same as after the commit, it just tricks the (not perfect) static analyser of the compiler to not print the warning.
I noticed this leak using a different static analyser on another project that uses glad.

Luflosi:
> only if realloc had previously succeeded

That doesn't mean that it's impossible. And if it was impossible, why are we using realloc in the first place?

Dav1dde:
I noticed the exact same thing after I 'fixed' it, looked into it again and realized it's not actually a problem since realloc is never used to resize but just to allocate, which indeed is not perfect just haven't changed that yet (no real priority).

If I am wrong here please correct me, also will look into it again and most likely replace the realloc with malloc. 

The reason realloc is used here is because this array used to be global and never freed, I changed that but must have forgotten to change the call to malloc.

PedDavid:
When I did the change it seemed to me like the caller dealt with free when an error code was returned. I just took another quick glance to the code and it doesn't seem like it. I agree that this could be improved, any one up for the task? (I volunteer if no one else)

Dav1dde:
@PedDavid if you wanna look at it again go for it.

> When I did the change it seemed to me like the caller dealt with free when an error code was returned. I just took another quick glance to the code and it doesn't seem like it.

The caller takes care of freeing, but only if the call succeeded and then resets the pointer to null, this means that the first argument to the realloc call should always be null -> realloc can be replaced with malloc or calloc, since it never resizes it just allocates.

Luflosi:
> any one up for the task?

This should be as simple as replacing `realloc` with `malloc` and removing the temporary variable, right? If so, I can make a pull request. I'd like to add a check just to be sure that `exts_i` is always null before the allocation. What would be the best way to do that? `if (exts_i != NULL) return 0;` or something else? Or should I not have such a check?

Dav1dde:
> This should be as simple as replacing `realloc` with `malloc` and removing the temporary variable, right? 

Yes.

> I'd like to add a check just to be sure that `exts_i` is always null before the allocation. What would be the best way to do that? `if (exts_i != NULL) return 0;` or something else? Or should I not have such a check?

I would not bother with such a check, since this is not supposed to happen and if it actually happens due to a bug in a different place, I'd rather keep leaking a small amount of memory than break stuff. 

An alternative solution is to get rid of the global variable all together (as it is done in [glad2](https://github.com/Dav1dde/glad/blob/glad2/glad/generator/c/templates/gl.c#L138)), but this is not required.

PR #181, opened 2018-10-20
Add an SDL2 sample for SDL users

rafaelcn:


Dav1dde:
Thank you!

I changed a few minor things:

* renamed the file to `sdl.c`
* removed the `#include` show-casing (I think this is just confusing, I should also remove it from the other example at one point)
* added the commandline options to build it
* prefixed the commit with "example:"

Pushed in: 26d0d4f

Dav1dde:
> Just a suggestion: the next time that you want some modifications to occur concerning the code please review the code and ask for them. I'd be glad to do it. 

I appreciate that, but I feel like with all these minor things I can just do it myself without bothering you again, since I usually checkout the code and test it anyways.

> One question, on [26d0d4f](https://github.com/Dav1dde/glad/commit/26d0d4f92d52c2a1c1980d94c1d47c4e66fb9c90) you commented on the second line the build command, but I think it needs to be compiled against `OpenGL` as well, hence `-lGL` right?

glad doesn't require to be linked with OpenGL in fact it's actually wrong to do that. If you do link with OpenGL and not you're not using any of the symbols the linker is just gonna ignore it. 
I don't know if there is a SDL2 setup that requires linking with OpenGL, on my system it is not necessary.

rafaelcn:
> I appreciate that, but I feel like with all these minor things I can just do it myself without bothering you again, since I usually checkout the code and test it anyways.

Right 😄 

> glad doesn't require to be linked with OpenGL in fact it's actually wrong to do that. If you do link with OpenGL and not you're not using any of the symbols the linker is just gonna ignore it.
I don't know if there is a SDL2 setup that requires linking with OpenGL, on my system it is not necessary.

Right, I understand that. I'm also not sure if it does needs to be linked with OpenGL when using SDL2. I gonna check that later and give you a feedback on that. Thank you once again for the information.

PR #191, opened 2018-12-02
Replace realloc with malloc

Luflosi:
As discussed in #178.


Dav1dde:
Thanks, looks good on first impression, I'll look into and test it tomorrow.

Dav1dde:
Thank you! Pushed in 8dcc73c

PR #199, opened 2019-02-06
Added support for a macro disabling the windows header

ImOnALampshade:
Users can define `GLAD_NO_INCLUDE_WINDOWS` before including glad.h in order to disable the Windows header from being included at all. If this macro is not present, behavior is unchanged, `Windows.h` is included with `WIN32_LEAN_AND_MEAN` and `NOMINMAX`.

### DIFF
--- a/glad/lang/c/loader/gl.py
+++ b/glad/lang/c/loader/gl.py
@@ -134,13 +134,23 @@ _OPENGL_HEADER_INCLUDE_ERROR = '''
 
 _OPENGL_HEADER = '''
 #if defined(_WIN32) && !defined(APIENTRY) && !defined(__CYGWIN__) && !defined(__SCITECH_SNAP__)
+
+#ifdef GLAD_NO_INCLUDE_WINDOWS
+#define APIENTRY __stdcall
+
+#else
+
 #ifndef WIN32_LEAN_AND_MEAN
 #define WIN32_LEAN_AND_MEAN 1
 #endif
 #ifndef NOMINMAX
 #define NOMINMAX 1
 #endif
+
 #include <windows.h>
+
+#endif
+
 #endif
 
 #ifndef APIENTRY


Dav1dde:
Why not just define `APIENTRY`? Is there a benefit here I don't see?

ImOnALampshade:
Do you mean `APIENTRYP`?  Or `GLAPIENTRY`?  Defining `APIENTRY` is exactly what the change does.

Either way, the answer is the same - I am not well versed in this project and didn't even learn about it until earlier today. I do not know if there would be wider implications caused by redefining either of those macros. I tried to tread as lightly as possible.

Dav1dde:
Sorry, I worded that badly.

If you define APIENTRY before including glad, windows.h won't be included:
```c
#ifdef _WIN32
    #define APIENTRY __stdcall
#endif

#include <glad.h>
```

Why is your change "better" than doing this?

ImOnALampshade:
My personal vendetta against `Windows.h` is that for the project I am using glad for, including it will dramatically increase (+50-75%) compile times, and silence our runtime error messages (Because it defines `ERROR` over the top of the macro we use). I also work entirely in unixy OS's, but others I work with use Windows. That means that when I drag in a new dependency that includes `Windows.h`, it causes a lot of problems.

Beyond my own personally issues with `Windows.h`, there's 3 big reasons I think this behavior is better:

# Reason 1

The only way to make sure that `APIENTRY` is defined whenever glad is included is to set a build rule for it - which will cause macro redefinition warnings if you include `Windows.h` anywhere else in your code (Such as in `glad.c`, which needs to include it).

If you were to do the same by defining `GLAD_NO_INCLUDE_WINDOWS` in your build system, then you would only get that warning if you included both `glad.h` and `Windows.h` in the same file, and also included `glad.h` *before* `Windows.h`. 

# Reason 2

It's more explicit for a user of glad to define `GLAD_NO_INCLUDE_WINDOWS` than to include `APIENTRY`.

The current behavior is that as a consequence of `APIENTRY` being defined, `Windows.h` is not included. This exposes irrelevant implementation details to the users of glad.

The change I'm proposing is to wrap that behavior into a macro that makes its consequence clear - `GLAD_NO_INCLUDE_WINDOWS` causes glad to not include `Windows.h`.

# Reason 3

Isn't the point of using a library like glad (Or glew, or glfw, etc) to remove platform specific code?

Dav1dde:
I agree including `windows.h` is unfortunate, even with all the "minification" defines.

Regarding 2) I agree an explicit define is prettier

Regarding 3) yes, but that's also the reason I include windows.h, it's supposed to have the "correct" value for `APIENTRY` - Is there a windows platform that does not define `APIENTRY` to `stdcall` (cygwin, windows rt, UWP)? If we "automatically" define `APIENTRY` I wanna be sure it works on non-default platforms as well (cygwin, windows rt, UWP etc.).

ImOnALampshade:
`APIENTRY` is always `__stdcall` (Or `WINAPI`, which is also `__stdcall`).

Here's `Windef.h` from [MinGW](https://github.com/Alexpux/mingw-w64/blob/master/mingw-w64-tools/widl/include/windef.h#L169), and a page about data types on [MSDN](https://docs.microsoft.com/en-us/windows/desktop/winprog/windows-data-types) (Ctrl + F for APIENTRY).

MinGW, Win32 apps, UWP's, or whatever else there is, all have to reference the same binaries to load functions out of - and need the calling convention set properly.

And for the purposes of what it's needed for in glad, no matter what compiler/header/sdk version you're compiling glad with, the DLL that the OpenGL functions are loaded out of is going to use `__stdcall`.

ImOnALampshade:
It's also worth noting that headers for other OpenGL loaders (Such as `glew`) define `APIENTRY` as `__stdcall` without issue.

Dav1dde:
Good to know, thanks for your research.

ImOnALampshade:
Following up - were there any other changes you wanted to make to this pull request (IE, do you want this functionality implemented differently), or would you like this PR abandoned?

Dav1dde:
I think this is a good functionality to have, thanks for your work and research on it. 

I am currently not sure about the name of the define, I don't want to have a clear naming schema on (potential future) defines also `GLAD_NO_INCLUDE_WINDOWS` "feels" wrong to me. I am also considering making this the default behaviour and completely removing the windows include (input welcome!)

Regarding the PR itself, I usually rebase smaller PRs manually and fix smaller things like commit naming myself, what I'd do here:

* prefix the commit message with `c: `
* remove superfluous newlines and indet the if/else instead

Dav1dde:
Took me a while, glad doesn't include `windows.h` for `APIENTRY` anymore.

mikejsavage:
I get `warning C4005: 'APIENTRY': macro redefinition` with this, rearranging glad.c a little bit fixes it:

```c
#if defined(_WIN32) || defined(__CYGWIN__)
#include <windows.h>
#endif

#include "glad.h"

static void* get_proc(const char *namez);

#if defined(_WIN32) || defined(__CYGWIN__)
static HMODULE libGL;
```

Dav1dde:
Thanks, this doesn't happen for me but is definitely possible, can you check if it's fixed now (or just add 
`#undef APIENTRY`)?

mikejsavage:
That fixes the warning too

Dav1dde:
Thanks again, I'll release a new version.

The reason for the undef is just that it is a lot easier to fix within glad (the includes get generated at a different place then the windows section).

PR #201, opened 2019-02-07
acknowledge MIT license per LICENSE file

haneefmubarak:


Dav1dde:
The code of glad (mainly Python) is licensed MIT the code that is generated by it is licensed public domain.

PR #214, opened 2019-06-02
obtaining EGL/eglplatform.h from a more reliable source

mchiasson:
in case nobody noticed, generating EGL from Glad2's web service (https://gen.glad.sh) gives us 504 Gateway Time-out. I deduced that the cause of this time out is from this link which also times out:

https://cgit.freedesktop.org/mesa/mesa/plain/include/EGL/eglplatform.h

I changed it to a more reliable and accurate source: https://www.khronos.org/registry/EGL/api/EGL/eglplatform.h

and it doesn't time out when I run glad2 from the command line at least.  As for the web service, let's ho[e that this rectifies the issue as I have no method of testing it to see if that is the only issue.

### DIFF
--- a/glad/generator/c/__init__.py
+++ b/glad/generator/c/__init__.py
@@ -185,7 +185,7 @@ class CGenerator(JinjaGenerator):
         Header(
             'eglplatform',
             'EGL/eglplatform.h',
-            'https://cgit.freedesktop.org/mesa/mesa/plain/include/EGL/eglplatform.h'
+            'https://www.khronos.org/registry/EGL/api/EGL/eglplatform.h'
         ),
         Header(
             'vk_platform',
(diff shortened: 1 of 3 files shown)


Dav1dde:
The reason glad queries `eglplatform.h` from freedesktop is because the khronos one was missing symbols. I'll have to see if Khronos updated their version.

Also this should not be happening, glad-web is supposed to include a cached copy not download the file on every request.

Dav1dde:
Thanks for the report, I temporarily fixed this on the server.

The better URL to use would be:
https://raw.githubusercontent.com/KhronosGroup/EGL-Registry/master/api/EGL/eglplatform.h

The old Khronos Url links to the old (unused) SVN iirc. If you want to update the PR please prefix the commit with `c:`.

PR #223, opened 2019-07-22
add CMake option GLAD_OUT_DIR

irov:
Add the ability to change the output installation path


Dav1dde:
Thanks for the PR, can you rebase those two commits into one and prefix the message with cmake?

I'll probably get around to test it sometime this week, I can do it then as well.

Dav1dde:
This is not working, see: https://cmake.org/cmake/help/v3.13/command/option.html

An option can either be ON/OFF. To make it configurable you have to make the variable cacheable instead. I messed around with it a bit, I don't really see a reason why to have it configurable, maybe you have a usecase for me.

irov:
Well, you are right. I create new pull request with fix this. ['option' it's way work, but don't standart]

PR #224, opened 2019-07-29
fix CMake variable GLAD_OUT_DIR

irov:
https://stackoverflow.com/questions/8709877/cmake-string-options


Dav1dde:
What is your usecase for this, why do you need the glad output in a different directory?

irov:
i use CI for build engine, and use some like this code:
```
        set(GLAD_INSTALL ON CACHE BOOL "GLAD_INSTALL" FORCE)
        set(USE_MSVC_RUNTIME_LIBRARY_DLL OFF CACHE BOOL "GLAD_INSTALL" FORCE)
        set(GLAD_OUT_DIR "${THIRDPARTY_DIR}/glad" CACHE STRING "GLAD_OUT_DIR" FORCE)

        ADD_SUBDIRECTORY(${THIRDPARTY_DIR}/glad ${CMAKE_TEMP_DIR}/glad)
```
if i don't setup GLAD_OUT_DIR - generated 'includes' locate in CMake temporary directory [${CMAKE_CURRENT_BINARY_DIR}]

Dav1dde:
Thanks!

PR #232, opened 2019-09-05
cmake: support for generating all extensions, fixes #230

### LINKED ISSUE #230: Impossible to specify all extensions using CMake
The documentation claims that it should be possible not to specify (a file containing ) a list of extensions for which to generate code.

However, `CMakeLists.txt` always passes `--extensions=${GLAD_EXTENSIONS}` to the Python program. As far as I understand, this makes it impossible to use cmake and not have to specify the list of extensions.

jackrabbit128:
Add a new option GLAD_ALL_EXTENSIONS, which is OFF by default. When turned ON, the value of the variable GLAD_EXTENSIONS is not passed on to the python code generator. As a result, the generated code includes all extensions in the requested APIs.

This works around the fact that there does not appear to be a way in CMake to distinguish an unset variable from one with an empty value. An empty GLAD_EXTENSIONS value indicates that the user does not want any extensions to be included.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,7 +23,7 @@ set(GLAD_API "" CACHE STRING "API type/version pairs, like \"gl=3.2,gles=\", no 
 set(GLAD_GENERATOR "c" CACHE STRING "Language to generate the binding for")
 set(GLAD_EXTENSIONS "" CACHE STRING "Path to extensions file or comma separated list of extensions, if missing all extensions are included")
 set(GLAD_SPEC "gl" CACHE STRING "Name of the spec")
-option(GLAD_ALL_EXTENSIONS "Include all extensions iso those specified by GLAD_EXTENSIONS" OFF)
+option(GLAD_ALL_EXTENSIONS "Include all extensions instead of those specified by GLAD_EXTENSIONS" OFF)
 option(GLAD_NO_LOADER "No loader" OFF)
 option(GLAD_REPRODUCIBLE "Reproducible build" OFF)
 if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -23,6 +23,7 @@ set(GLAD_API "" CACHE STRING "API type/version pairs, like \"gl=3.2,gles=\", no
 set(GLAD_GENERATOR "c" CACHE STRING "Language to generate the binding for")
 set(GLAD_EXTENSIONS "" CACHE STRING "Path to extensions file or comma separated list of extensions, if missing all extensions are included")
 set(GLAD_SPEC "gl" CACHE STRING "Name of the spec")
+option(GLAD_ALL_EXTENSIONS "Include all extensions iso those specified by GLAD_EXTENSIONS" OFF)
```
Dav1dde: Typo in the description?

Dav1dde:
Thanks for the change, that's pretty much what I had in mind!

CMakeLists.txt
jackrabbit128: Idiosyncrasy of mine: iso = instead of. Will adjust

jackrabbit128:
I am not that familiar with the github flow. Hopefully I did what you expected.

Dav1dde:
Thanks I pushed it in https://github.com/Dav1dde/glad/commit/2a5ac49ba7b5275e6f0f6d1fa07bfb0e6875ce18

What I meant was to rename the git commit, you can do this using `git commit --amend` or an interactive rebase `git rebase -i HEAD~X`. The reason is I try to have a relatively clean commit history which helps in the wrong run if you want to find out why this line was changed and in what context the change happened.

I went ahead and did that for you, the changes are now in master. Thanks again for the PR!

PR #237, opened 2019-12-10
glad: Fix EGL loading for EGL_NO_DISPLAY on EGL 1.5 and above

thochstein:
EGL loading was failing for a display of EGL_NO_DISPLAY.
Allow loading EGL 1.5 and above since eglQueryString supports
EGL_NO_DISPLAY in this case.


Dav1dde:
Thanks makes sense after reading the docs. Gotta look a bit more into it though (does calling `eglQueryString` with no display return less extensions - I assume it does?)

PR #256, opened 2020-03-14
glad2 Rust Multi-Context

### LINKED ISSUE #254: glad2 Rust Multicontext
As far as I can see, multicontext is implemented and working for C/C++ in glad2; something similar should be possible for Rust as well. (see: [`gl_generator` "Struct Generator"](https://github.com/brendanzab/gl-rs/tree/master/gl_generator))

I am unfamiliar with glad's internals, but this like it would be pretty trivial to implement—at first glance. If no immediate work is being done/is planned, I *might* be able to eventually take a go myself; nevertheless, opening this issue because I feel like this is something that should be implemented eventually.

solarane:
Hopefully fixes #254.

It's 6 AM right now. I worked on this through the night. Probably not the best idea.
Some things to note:
- I only performed basic sanity checks; integration tests are needed
- ~~Somehow, I ended up changing the order of certain aliases even without the `--mx` flag. I'm 99% sure I just changed the order but again, integration tests are needed~~
- The formatting of `impl.rs` is a bit weird. It was a bit weird before I laid my hands on it, and it is probably weirder now because I've never used `jinja2` before. Some code review would probably be good.
- ~~Currently, with `--mx`, I don't cache function pointers. This means that for every function with an alias, for every alias, a new function pointer is created (although, of course, they are then immediately dropped). This should probably be addressed.~~

Now I'm going to go get some sleep. If y'all see something wrong, please let me know.

(CC: @Dav1dde)

### DIFF
--- a/glad/generator/rust/templates/impl.rs
+++ b/glad/generator/rust/templates/impl.rs
@@ -12,6 +12,16 @@ pub struct FnPtr {
 }
 
 impl FnPtr {
+    {% if options.mx %}
+    pub fn new<F>(loadfn: &mut F, name: &'static str) -> FnPtr where F: FnMut(&'static str) -> *const raw::c_void {
+        let loaded = loadfn(name);
+        if !loaded.is_null() {
+            FnPtr { ptr: loaded, is_loaded: true }
+        } else {
+            FnPtr { ptr: FnPtr::not_initialized as *const raw::c_void, is_loaded: false }
+        }
+    }
+    {% else %}
     pub fn empty() -> FnPtr {
         FnPtr { ptr: FnPtr::not_initialized as *const raw::c_void, is_loaded: false }
     }
@@ -26,26 +36,14 @@ impl FnPtr {
             self.is_loaded = false;
         };
     }
-
+    
+    {% endif %}
     pub fn aliased(&mut self, other: &FnPtr) {
         if !self.is_loaded && other.is_loaded {
             self.ptr = other.ptr;
             self.is_loaded = other.is_loaded;
         }
     }
-    {% if options.mx %}
-    pub fn init_load_aliased<F>(loadfn: &mut F, name: &'static str, others: &[&'static str]) -> Self where F: FnMut(&'static str) -> *const raw::c_void {
-        let mut ptr_a = Self::empty();
-        ptr_a.load(loadfn, name);
-        for other in others {
-            let mut ptr_b = Self::empty();
-            ptr_b.load(loadfn, other);
-            ptr_a.aliased(&ptr_b);
-        }
-
-        ptr_a
-    }
-    {% endif %}
 
     #[inline(never)]
     fn not_initialized() -> ! { panic!("{{ feature_set.name }}: function not initialized") }
@@ -77,7 +75,7 @@ pub mod functions {
     {{ 'pub struct Gl {' if options.mx }}
     {% if options.mx %}
     {% for command in feature_set.commands %}
-    {{ '    ' }}{{ template_utils.protect(command) }} pub _{{ command.name|no_prefix }}: FnPtr,
+    {{ '    ' }}{{ template_utils.protect(command) }} pub(super) _{{ command.name|no_prefix }}: FnPtr,
     {% endfor %}
     }
 
@@ -104,23 +102,27 @@ mod storage {
 
 pub fn load<F>(mut loadfn: F) {{ '-> functions::Gl' if options.mx }} where F: FnMut(&'static str) -> *const raw::c_void {
     {% if options.mx %}
-    Gl {
+    let mut gl = Gl {
     {% for command in feature_set.commands %}
-    {{ '    ' }}{{ template_utils.protect(command.name) }} _{{ command.name|no_prefix }}: FnPtr::init_load_aliased(&mut loadfn, "{{ command.name }}", &[{% for alias in aliases.get(command.name)|reject('equalto', command) %}"{{ alias }}", {% endfor %}]),
+    {{ '    ' }}{{ template_utils.protect(command.name) }} _{{ command.name|no_prefix }}: FnPtr::new(&mut loadfn, "{{ command.name }}"),
     {% endfor %}
-    }
+    };
     {% else %}
     unsafe {
         {% for command in feature_set.commands %}
         {{ template_utils.protect(command) }} storage::{{ command.name | no_prefix }}.load(&mut loadfn, "{{ command.name }}");
         {% endfor %}
-
-        {% for command, caliases in aliases|dictsort %}
-        {% for alias in caliases|reject('equalto', command) %}
-        {{ template_utils.protect(command) }} storage::{{ command|no_prefix }}.aliased(&storage::{{ alias|no_prefix }});
-        {% endfor %}
-        {% endfor %}
     }
+    {% endif %}
+
+    {% for command, caliases in aliases|dictsort %}
+    {% for alias in caliases|reject('equalto', command) %}
+    {{ template_utils.protect(command) }} {{ 'gl._' if options.mx else 'storage::' }}{{ command|no_prefix }}.aliased(&{{ 'gl._' if options.mx else 'storage::' }}{{ alias|no_prefix }});
+    {% endfor %}
+    {% endfor %}
+    {% if options.mx %}
+     
+     gl
     {% endif %}
 }
 
--- a/glad/generator/rust/templates/impl.rs
+++ b/glad/generator/rust/templates/impl.rs
@@ -5,14 +5,18 @@ pub use self::functions::*;
 
 use std::os::raw;
 
+{% if options.mx %}
+struct FnPtr {
+{% else %}
 pub struct FnPtr {
+{% endif %}
     ptr: *const raw::c_void,
     is_loaded: bool
 }
 
 impl FnPtr {
     {% if options.mx %}
-    pub fn new<F>(loaded: *const raw::c_void) -> FnPtr where F: FnMut(&'static str) -> *const raw::c_void {
+    pub fn new(loaded: *const raw::c_void) -> FnPtr {
         if !loaded.is_null() {
             FnPtr { ptr: loaded, is_loaded: true }
         } else {
@@ -36,6 +40,7 @@ impl FnPtr {
     }
     
     {% endif %}
+    #[allow(dead_code)]
     pub fn aliased(&mut self, other: &FnPtr) {
         if !self.is_loaded && other.is_loaded {
             self.ptr = other.ptr;
@@ -100,8 +105,9 @@ mod storage {
 {% endif %}
 
 {% if options.mx %}
+#[allow(unused_mut)]
 pub fn load<F>(mut loadfn: F) -> functions::Gl where F: FnMut(&'static str) -> *const raw::c_void {
-    let gl = Gl {
+    let mut gl = Gl {
         {% for command in feature_set.commands %}
         {{ template_utils.protect(command.name) }} _{{ command.name|no_prefix }}: FnPtr::new(loadfn("{{ command.name }}")),
         {% endfor %}
--- a/glad/generator/rust/templates/impl.rs
+++ b/glad/generator/rust/templates/impl.rs
@@ -1,5 +1,4 @@
 {% import 'template_utils.rs' as template_utils with context %}
-#![allow(non_snake_case)]
 pub use self::types::*;
 pub use self::enumerations::*;
 pub use self::functions::*;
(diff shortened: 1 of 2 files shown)


glad/generator/rust/templates/impl.rs
```diff
@@ -1,5 +1,5 @@
 {% import 'template_utils.rs' as template_utils with context %}
-
+{{ '#![allow(non_snake_case)]' if options.mx }}
```
Dav1dde: Just allows allow, makes the template more readable and doesn't matter in the generated code

glad/generator/rust/templates/impl.rs
```diff
@@ -54,20 +67,31 @@ pub mod enumerations {
 }
 
 pub mod functions {
-    #![allow(non_snake_case, unused_variables, dead_code)]
+    #![allow({{ 'non_snake_case, ' if not options.mx }}unused_variables, dead_code)]
```
Dav1dde: Same here and all the other's just allow what you have to allow

glad/generator/rust/templates/impl.rs
```diff
@@ -54,20 +67,31 @@ pub mod enumerations {
...
     use std::mem;
-    use super::storage;
+    {{ 'use super::FnPtr;' if options.mx else 'use super::storage;' }}
     use super::types::*;
 
+    {{ 'pub struct Gl {' if options.mx }}
+    {% if options.mx %}
     {% for command in feature_set.commands %}
-    {{ template_utils.protect(command) }} #[inline] pub unsafe fn {{ command.name|no_prefix }}({{ command|params }}) -> {{ command.proto.ret|type }} { mem::transmute::<_, extern "system" fn({{ command|params('types') }}) -> {{ command.proto.ret|type }}>(storage::{{ command.name|no_prefix }}.ptr)({{ command|params('names') }}) }
+    {{ '    ' }}{{ template_utils.protect(command) }} pub _{{ command.name|no_prefix }}: FnPtr,
```
Dav1dde: `{{ '    ' }}` can just indent by that much ;)

glad/generator/rust/templates/impl.rs
```diff
@@ -76,8 +100,16 @@ mod storage {
     {{ template_utils.protect(command) }} pub static mut {{ command.name|no_prefix }}: FnPtr = FnPtr { ptr: FnPtr::not_initialized as *const raw::c_void, is_loaded: false };
     {% endfor %}
 }
+{% endif %}
 
-pub fn load<F>(mut loadfn: F) where F: FnMut(&'static str) -> *const raw::c_void {
+pub fn load<F>(mut loadfn: F) {{ '-> functions::Gl' if options.mx }} where F: FnMut(&'static str) -> *const raw::c_void {
```
Dav1dde: I feel like having to separate loader functions would be more readable:

```
if mx:
  define load() -> GL
else:
  define load()
```

Dav1dde:
Looks like I submitted my review the same minute you updated the branch. Just ignore what's outdated, it's just style stuff anyways.

solarane:
> Currently, with `--mx`, I don't cache function pointers. This means that for every function with an alias, for every alias, a new function pointer is created (although, of course, they are then immediately dropped). This should probably be addressed.

Fixed. Bindings generated with `--mx` is still 10 lines shorter than those generated without it.

> Somehow, I ended up changing the order of certain aliases even without the `--mx` flag. I'm 99% sure I just changed the order but again, integration tests are needed

Upon closer inspection, it actually seems that the order of certain aliases is nondeterministic. Not sure if this is intended behavior or a bug.

Anyways, both issues with the original commit are now fixed.

Dav1dde:
> > Somehow, I ended up changing the order of certain aliases even without the `--mx` flag. I'm 99% sure I just changed the order but again, integration tests are needed
> 
> Upon closer inspection, it actually seems that the order of certain aliases is nondeterministic. Not sure if this is intended behavior or a bug.

A "bug", looks like the [function generating the aliases](https://github.com/Dav1dde/glad/blob/glad2/glad/generator/util.py#L48) is returning a dict, should probably be some kind of sorted/ordered dictionary.

Can ignore this for now, I'll fix this separately.

glad/generator/rust/templates/impl.rs
solarane: Makes sense :) Should be fixed in https://github.com/Dav1dde/glad/pull/256/commits/3c1d7934900137cecf580409a64b890e30e97a1c.

glad/generator/rust/templates/impl.rs
solarane: Should be fixed in https://github.com/Dav1dde/glad/pull/256/commits/3c1d7934900137cecf580409a64b890e30e97a1c.

glad/generator/rust/templates/impl.rs
solarane: Heh, cool! Didn't know that :D Fixed.

glad/generator/rust/templates/impl.rs
solarane: I assume you mean just the function signature? If so, fixed in https://github.com/Dav1dde/glad/pull/256/commits/3c1d7934900137cecf580409a64b890e30e97a1c. Think it's definitely more readable now.

solarane:
> Can ignore this for now, I'll fix this separately.
Gotcha.

Integration tests were failing (non-`mx`) because of a small oversight (see https://github.com/Dav1dde/glad/pull/256/commits/3c1d7934900137cecf580409a64b890e30e97a1c), now fixed. Style things you mentioned should also be fixed. Thanks for the quick feedback!

glad/generator/rust/templates/impl.rs
Dav1dde: Yeah the function signature or even the whole function, since you're basically splitting the body into two separate parts anyways.

glad/generator/rust/templates/impl.rs
```diff
@@ -76,7 +98,25 @@ mod storage {
     {{ template_utils.protect(command) }} pub static mut {{ command.name|no_prefix }}: FnPtr = FnPtr { ptr: FnPtr::not_initialized as *const raw::c_void, is_loaded: false };
     {% endfor %}
 }
+{% endif %}
 
+{% if options.mx %}
+pub fn load<F>(mut loadfn: F) -> functions::Gl where F: FnMut(&'static str) -> *const raw::c_void {
+    let mut gl = Gl {
```
Dav1dde: doesn't need to be mut

glad/generator/rust/templates/impl.rs
```diff
@@ -1,5 +1,5 @@
 {% import 'template_utils.rs' as template_utils with context %}
-
+#![allow(non_snake_case)]
```
Dav1dde: From my tests this is not necessary, and if it was we should move it to the block where it is required (as with the other allows)

glad/generator/rust/templates/impl.rs
```diff
@@ -54,20 +65,31 @@ pub mod enumerations {
...
+        {{ template_utils.protect(command) }} pub(super) _{{ command.name|no_prefix }}: FnPtr,
+        {% endfor %}
+    }
+
+    impl Gl {
+    {% endif %}
+        {% for command in feature_set.commands %}
+        {{ template_utils.protect(command) }} #[inline] pub unsafe fn {{ command.name|no_prefix }}({{ '&self, ' if options.mx }}{{ command|params }}) -> {{ command.proto.ret|type }} { mem::transmute::<_, extern "system" fn({{ command|params('types') }}) -> {{ command.proto.ret|type }}>({{ 'self._' if options.mx else 'storage::' }}{{ command.name|no_prefix }}.ptr)({{ command|params('names') }}) }
+        {% endfor %}
+    {{ '}' if options.mx }}
```
Dav1dde:
```rust
    {% if options.mx %}
    pub struct Gl {
        {% for command in feature_set.commands %}
        {{ template_utils.protect(command) }} pub(super) _{{ command.name|no_prefix }}: FnPtr,
        {% endfor %}
    }
    impl Gl {
    {% endif %}

    {% for command in feature_set.commands %}
    {{ template_utils.protect(command) }} #[inline] pub unsafe fn {{ command.name|no_prefix }}({{ '&self, ' if options.mx }}{{ command|params }}) -> {{ command.proto.ret|type }} { mem::transmute::<_, extern "system" fn({{ command|params('types') }}) -> {{ command.proto.ret|type }}>({{ 'self._' if options.mx else 'storage::' }}{{ command.name|no_prefix }}.ptr)({{ command|params('names') }}) }
    {% endfor %}

    {{ '}' if options.mx }}
```

An idea of making it a bit more readble, clearly visible which block is the `if options.mx` block and one less template if

glad/generator/rust/templates/impl.rs
```diff
@@ -54,20 +65,31 @@ pub mod enumerations {
...
 }
 
+{% if not options.mx %}
 mod storage {
-    #![allow(non_snake_case, non_upper_case_globals)]
```
Dav1dde: Still need this if you remove the global allow

glad/generator/rust/templates/impl.rs
```diff
@@ -12,6 +12,16 @@ pub struct FnPtr {
 }
 
 impl FnPtr {
+    {% if options.mx %}
+    pub fn new<F>(loadfn: &mut F, name: &'static str) -> FnPtr where F: FnMut(&'static str) -> *const raw::c_void {
```
Dav1dde: What about passing the `raw::c_void` as argument instead of the loader and the name (the loading logic shouldn't be associated with the `FnPtr` imo).

```rust
FnPtr::new(load("glClear"))
```

Dav1dde:
Thanks for the work! I just noted a few minor things rest looks good.

Also reminded me that I have to give Rust a little bit more love, clean up the visibility (there is way too much pub) and use `FnPtr::empty()` instead of initializing manually on every function...

Dav1dde:
@solarane if you want I can take care of the last minor changes

solarane:
@Dav1dde oh god, so sorry. I initially had this on my list of TODOs for later, but it completely slipped out of my mind D: I will get to it tomorrow (it's 2 AM where I am currently) and resolve your reviews. Sorry again.

Dav1dde:
No worries take your time

glad/generator/rust/templates/impl.rs
solarane: Indents will be wrong this way, but shouldn't be an issue. definitely much more readable

solarane:
@Dav1dde All your reviews should now be addressed. Tests are passing. Ready for merge?

glad/generator/rust/templates/impl.rs
```diff
@@ -12,6 +11,15 @@ pub struct FnPtr {
 }
 
 impl FnPtr {
+    {% if options.mx %}
+    pub fn new<F>(loaded: *const raw::c_void) -> FnPtr where F: FnMut(&'static str) -> *const raw::c_void {
```
Dav1dde, line 15: `F` is not used anymore and leads to compiler errors "cannot infer F"

glad/generator/rust/templates/impl.rs
```diff
@@ -76,7 +97,25 @@ mod storage {
     {{ template_utils.protect(command) }} pub static mut {{ command.name|no_prefix }}: FnPtr = FnPtr { ptr: FnPtr::not_initialized as *const raw::c_void, is_loaded: false };
     {% endfor %}
 }
+{% endif %}
+
+{% if options.mx %}
+pub fn load<F>(mut loadfn: F) -> functions::Gl where F: FnMut(&'static str) -> *const raw::c_void {
+    let gl = Gl {
```
Dav1dde, line 104: I told you this doesn't need to be mut, sorry in fact it has to be `mut` because of the later `aliased` calls, but we need to silence the warning in case there are no aliased functions

glad/generator/rust/templates/impl.rs
```diff
@@ -26,7 +34,8 @@ impl FnPtr {
             self.is_loaded = false;
         };
     }
-
+    
+    {% endif %}
     pub fn aliased(&mut self, other: &FnPtr) {
```
Dav1dde, line 39: It is possible that there are no aliased functions, then this is dead code, the struct or function needs `#[allow(dead_code)]`

Dav1dde:
@solarane I am testing the review feature (never really used it before), I think you get that changes are requested because I have to accept the first?

Anyways there are a few things needed in order for it to compile, I pushed an example (and simple test) to the [rust-mx-add](https://github.com/Dav1dde/glad/tree/rust-mx-add) branch. If you're using it just make sure to generate once without extensions (default) and with all extensions (for no aliasing and aliasing).

solarane:
@Dav1dde All right, all should be fixed.

In addition to stuff you reviewed, I saw that you changed `pub struct FnPtr` to `pub(crate)`. Since it's at the root of the crate, I'm pretty sure that's just equivalent to not making it `pub`, which is what I did. Also this time, I tested that everything compiles and runs as expected. :3

Dav1dde:
@solarane thank you for your work! I went ahead and rebased the commits and pushed them as 91c7599

solarane:
Thanks for the merge!
~~I didn't get the contributor badge, which is sad. Guess that's because this PR wasn't merged directly? D:~~
I've also noticed a bug (?) with how multiple contexts interact with multiple threads, which would require changing the internal representations of the function pointers. I'll open a new issue on that, and probably a PR sometime soon...

Dav1dde:
@solarane yeah I wish github wasn't like this, all my commits to glad2 do not count towards the statistics either.

Please do open an issue.

PR #267, opened 2020-04-20
Switch to Python3 preferred

jherico:
While working on a Vulkan / OpenGL interoperability [example](https://github.com/KhronosGroup/Vulkan-Samples/pull/97) I encountered across a [build failure](https://travis-ci.com/github/KhronosGroup/Vulkan-Samples/jobs/320793427) that appears to be related to Python2 usage in the Glad generator.  Since Python2 is finally deprecated, I thought it would be wise to switch to the updated CMake FindPython module that preferentially searches for Python3.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,16 @@ endif()
 project(GLAD VERSION 0.1.33 LANGUAGES C)
 
 set(GLAD_DIR "${CMAKE_CURRENT_SOURCE_DIR}")
-find_package(PythonInterp REQUIRED)
+
+# Find the python interpreter, set the PYTHON_EXECUTABLE variable
+if (CMAKE_VERSION VERSION_LESS 3.12)
+    # this logic is deprecated in CMake after 3.12
+    find_package(PythonInterp REQUIRED)
+else()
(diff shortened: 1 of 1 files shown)


Dav1dde:
Ah yeah it's definitely time to switch. I'll be supporting Python 2 for a lot longer though (no reason to drop it).

I'll go over this tomorrow

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.0)
+cmake_minimum_required(VERSION 3.12)
```
jherico, line 1: Minimum version for the new [FindPython](https://cmake.org/cmake/help/v3.12/module/FindPython.html) module in CMake

Dav1dde: Is there a possibility to check if the current cmake version is >= 3.12 then use the new FindPython, otherwise default to the old way (with maybe a warning)?

I don't want to break anyone's build

jherico: fixed

Dav1dde:
@jherico thanks for the contribution, I changed the commit message and pushed it manually as commit fd8ea1a

Issue #271, opened 2020-05-21
Vulkan structs from extensions typedef'ed incorrectly


nickclark2016:
When generating Vulkan headers with header only and loaders enabled, extensions that have been migrated to Vulkan Core have an incorrect struct typedef in the header.  
For example, the output is:
```cpp
typedef struct VkBindImageMemoryInfo {
    VkStructureType   sType;
    const  void *                       pNext;
    VkImage                            image;
    VkDeviceMemory                     memory;
    VkDeviceSize                       memoryOffset;
} VkBindImageMemoryInfo;
typedef struct VkBindImageMemoryInfoKHR  VkBindImageMemoryInfoKHR;
```
Expected output should be:
```cpp
typedef struct VkBindImageMemoryInfo {
    VkStructureType   sType;
    const  void *                       pNext;
    VkImage                            image;
    VkDeviceMemory                     memory;
    VkDeviceSize                       memoryOffset;
} VkBindImageMemoryInfo;
typedef struct VkBindImageMemoryInfo  VkBindImageMemoryInfoKHR;
```
This example above can be reproduced with the VK_KHR_bind_memory2 extension, which is required for VMA.

Dav1dde:
Thank you! Pushed in e8a8dd3

PR #272, opened 2020-05-21
c: Fix generation of struct aliases

nickclark2016:
This commit fixes the issue presented by #271 by checking if the struct has an alias. If so, the alias field should be the struct name, rather than the name field of the spec.


Dav1dde:
Pushed in e8a8dd3 - I just removed the `.vscode` config and added the github references to the commit.

Thanks!

nickclark2016:
Thanks for looking at this so quickly!

PR #276, opened 2020-06-13
Added support for building multiple specs from a single CMakeLists.txt

handicraftsman:
All specs are built in a single object/archive file at the moment


Dav1dde:
I am not a fan of the split between `GLAD_SPEC` and `GLAD_ADDITIONAL_SPECS`, is it possible to just use `GLAD_SPEC` in a backwards compatible manner?

handicraftsman:
Didn't check, this change was done as a patch for vcpkg's glad port first. Although yes, there should be a better way to do this.

Dav1dde:
Obsolete now because of #316

PR #284, opened 2020-06-24
Fix CMake IPO warning

AMS21:
If you enable IPO in your own CMakeLists.txt and then include glad via add_subdirectory like this.
```cmake
include(CheckIPOSupported)
check_ipo_supported(RESULT ipo_result OUTPUT ipo_output)
if(ipo_result)
  set(CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE)
endif()
(...)
add_subdirectory(extlib/glad)
```
Your going to get the following warning from CMake starting from version 3.9:
```
CMake Warning (dev) at extlib/glad/CMakeLists.txt:103 (add_library):
Policy CMP0069 is not set: INTERPROCEDURAL_OPTIMIZATION is enforced when
enabled.  Run "cmake --help-policy CMP0069" for policy details.  Use the
cmake_policy command to set the policy and suppress this warning.

INTERPROCEDURAL_OPTIMIZATION property will be ignored for target 'glad'.
This warning is for project developers.  Use -Wno-dev to suppress it.
```

This patch should fix that warning and allow glad to be build with link time optimizations with non Intel compilers like Clang or GCC, when enabled. Tested on my local machine using CMake 3.17.3.

[CMP0069 Refernce](https://cmake.org/cmake/help/git-stage/policy/CMP0069.html)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,6 +11,11 @@ if(NOT CMAKE_VERSION VERSION_LESS 3.1)
   cmake_policy(SET CMP0054 NEW)
 endif()
 
+if(CMAKE_VERSION VERSION_GREATER 3.8)
+  # Enable IPO for CMake versions that support it
+  cmake_policy(SET CMP0069 NEW)
+endif()
+
 project(GLAD VERSION 0.1.33 LANGUAGES C)
 
 set(GLAD_DIR "${CMAKE_CURRENT_SOURCE_DIR}")


Dav1dde:
Don't we have to check for support before enabling?

```
include(CheckIPOSupported)
check_ipo_supported()
```

AMS21:
Yes if you want build with IPO you need to enable it like this for example.
```cmake
include(CheckIPOSupported)
check_ipo_supported(RESULT ipo_result OUTPUT ipo_output)
if(ipo_result)
  set(CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE)
endif()
```
Maybe I didn't state the effects of my patch very clearly sorry for that. This patch does *not* generally enable IPO for glad or even add an option to directly enable it. The responsibility for enabling IPO still lies at the user of the library just like before.

The problem this patch fixes it that with CMake versions after 3.9 globally enabling IPO using the CMAKE_INTERPROCEDURAL_OPTIMIZATION would not work. That's what the warning message above is hinting at.

To quote the documentation on [Policy 0069](https://cmake.org/cmake/help/git-stage/policy/CMP0069.html)
> Since CMake 3.8 and lower only honored INTERPROCEDURAL_OPTIMIZATION for the Intel compiler on Linux, some projects may unconditionally enable the target property. Policy CMP0069 provides compatibility with such projects.

> This policy takes effect whenever the IPO property is enabled. The OLD behavior for this policy is to add IPO flags only for Intel compiler on Linux. The NEW behavior for this policy is to add IPO flags for the current compiler or produce an error if CMake does not know the flags.

Dav1dde:
Thanks for the explanation and contribution! 

I manually pushed this in df8e9e1, just changed the commit message.

PR #285, opened 2020-06-28
Undefine APIENTRY at the end of glad.h

### LINKED ISSUE #283: Including `Windows.h` after `glad.h` raises "macro redefinition" warning
When compiling the following code snippet on Windows:

```cpp
#include <glad/glad.h>
#include <Windows.h>

int main() {}
```

I get the following warning:

```
[...]\minwindef.h(130,1): warning C4005: 'APIENTRY': macro redefinition (compiling source file [...])
[...]\include\glad/glad.h(32): message : see previous definition of 'APIENTRY' (compiling source file [...])
```

This is with glad 0.1.33, using the webservice: https://glad.dav1d.de/#profile=core&language=c&specification=gl&api=gl%3D3.0

I have enabled warnings as errors in my project, which means I need to undefine `APIENTRY` after including `glad.h` in order to have a successful build. Undefining the macro at the end of `glad.h` would prevent that problem.

lethal-guitar:
Previously, including `Windows.h` after `glad.h` would trigger a warning due to `APIENTRY` being redefined by the former. Undefining `APIENTRY` prevents that problem.

Closes #283

### DIFF
--- a/glad/lang/c/loader/gl.py
+++ b/glad/lang/c/loader/gl.py
@@ -135,10 +135,12 @@ _OPENGL_HEADER_INCLUDE_ERROR = '''
 _OPENGL_HEADER = '''
 #if defined(_WIN32) && !defined(APIENTRY) && !defined(__CYGWIN__) && !defined(__SCITECH_SNAP__)
 #define APIENTRY __stdcall
+#define GLAD_DEFINED_APIENTRY
 #endif
 
 #ifndef APIENTRY
 #define APIENTRY
+#define GLAD_DEFINED_APIENTRY
 #endif
 #ifndef APIENTRYP
 #define APIENTRYP APIENTRY *
@@ -169,6 +171,11 @@ GLAPI int gladLoadGL(void);
 _OPENGL_HEADER_END = '''
 #ifdef __cplusplus
 }
+#endif
+
+#if defined(GLAD_DEFINED_APIENTRY)
+#undef APIENTRY
+#undef GLAD_DEFINED_APIENTRY
 #endif
 
 #endif


glad/lang/c/loader/__init__.py
```diff
@@ -3,9 +3,6 @@
 %(pre)s void* %(proc)s(const char *namez);
 
 #if defined(_WIN32) || defined(__CYGWIN__)
-#ifndef _WINDOWS_
```
Dav1dde, line 6: I think we still need this block, in case `APIENTRY` is already defined from somewhere that is not `windows.h`.

glad/lang/c/loader/gl.py
```diff
@@ -171,6 +171,10 @@
 }
 #endif
 
+#if !defined(_WINDOWS_)
```
Dav1dde, line 174: Instead of check for `_WINDOWS_` we should set a flag when we define `APIENTRY` maybe something like `GLAD_DEFINED_APIENTRY` and when this is defined `undef` both symbols `APIENTRY` and `GLAD_DEFINED_APIENTRY`.

lethal-guitar: Makes sense 👍

lethal-guitar:
@Dav1dde thanks for the quick feedback. I think I've addressed all your comments now, let me know if you'd like something else to be changed.

Dav1dde:
Mh unfortunately it is not that easy, your changes dont compile:

```c
In file included from src/glad.c:25:
include/glad/glad.h:41:28: error: expected ‘)’ before ‘*’ token
   41 | #define APIENTRYP APIENTRY *
      |                            ^
src/glad.c:83:16: note: in expansion of macro ‘APIENTRYP’
   83 | typedef void* (APIENTRYP PFNGLXGETPROCADDRESSPROC_PRIVATE)(const char*);
      |                ^~~~~~~~~
src/glad.c:84:8: error: unknown type name ‘PFNGLXGETPROCADDRESSPROC_PRIVATE’
   84 | static PFNGLXGETPROCADDRESSPROC_PRIVATE gladGetProcAddressPtr;
      |        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/glad.c: In function ‘open_gl’:
src/glad.c:108:38: error: ‘PFNGLXGETPROCADDRESSPROC_PRIVATE’ undeclared (first use in this function)
  108 |             gladGetProcAddressPtr = (PFNGLXGETPROCADDRESSPROC_PRIVATE)dlsym(libGL,
      |                                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/glad.c:108:38: note: each undeclared identifier is reported only once for each function it appears in
src/glad.c:108:71: error: expected ‘;’ before ‘dlsym’
  108 |             gladGetProcAddressPtr = (PFNGLXGETPROCADDRESSPROC_PRIVATE)dlsym(libGL,
      |                                                                       ^~~~~
      |                                                                       ;
src/glad.c:110:42: warning: comparison between pointer and integer
  110 |             return gladGetProcAddressPtr != NULL;
      |                                          ^~
src/glad.c: In function ‘get_proc’:
src/glad.c:133:30: warning: comparison between pointer and integer
  133 |     if(gladGetProcAddressPtr != NULL) {
      |                              ^~
src/glad.c:134:18: error: called object ‘gladGetProcAddressPtr’ is not a function or function pointer
  134 |         result = gladGetProcAddressPtr(namez);
      |                  ^~~~~~~~~~~~~~~~~~~~~
src/glad.c:84:41: note: declared here
   84 | static PFNGLXGETPROCADDRESSPROC_PRIVATE gladGetProcAddressPtr;
      |     
```

Thinking about it, it  makes a lot of sense, `APIENTRY` is used within macros which will get resolved *outside* of `glad.h` but outside of it `APIENTRY` does not exist anymore, because we `undef`'d it.

I can't think of an easy solution.

lethal-guitar:
Oh I see, I should have tested the case where a loader is generated. Doesn't seem so easy to fix indeed. Thanks for trying it out, I will have a look at the glad2 branch.

Dav1dde:
@lethal-guitar thanks for your work, sorry I didnt see this before. I hope glad2 fixes your issue, feel free to open an issue if you find another issue or need help.

PR #294, opened 2020-09-01
better error handling in glad_vk_get_extensions()

### LINKED ISSUE #293: [glad2] Possible memory leak in glad_vk_get_extensions()
In `glad_vk_get_extensions()`, there is an early return where `ext_properties` doesn't get freed.

In this part (`glad/generator/c/templates/vk.c`):
```
    if (physical_device != NULL) {
        result = {{ 'vkEnumerateDeviceExtensionProperties'|ctx }}(physical_device, NULL, &device_extension_count, ext_properties);
        if (result != VK_SUCCESS) {
            for (i = 0; i < instance_extension_count; ++i) {
                free((void*) extensions[i]);
            }
            free(extensions);
            return 0;
        }
```
There must be a `free((void*) ext_properties);` before the return.

I noticed #209 mentioned a few other warnings, all in this function. I believe you should be checking for non-zero return value from those `malloc`s, and maybe for non-zero `instance_extension_count + device_extension_count`.

What do you think about using `goto` for clearer error handling in this function?

gk7huki:
Fixes #293

### DIFF
--- a/glad/generator/c/templates/vk.c
+++ b/glad/generator/c/templates/vk.c
@@ -6,10 +6,10 @@ static int glad_vk_get_extensions({{ template_utils.context_arg(',') }} VkPhysic
     uint32_t i;
     uint32_t instance_extension_count = 0;
     uint32_t device_extension_count = 0;
-    uint32_t max_extension_count;
-    uint32_t total_extension_count;
-    char **extensions;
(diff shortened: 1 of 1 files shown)


Dav1dde:
Pushed in b405ebf - just changed the commit message.

Thank you for the contribution @gk7huki !

PR #299, opened 2020-09-20
Add vulkan nim generator

jacereda:
A quick attempt at a vulkan nim generator. The code is quite ugly but it seems to get the job done. I'll ask for feedback regarding the bindings to the nim community, so feel free to ignore this until the bindings are properly reviewed. Also, any comment on how to improve the code is welcome.

### DIFF
--- a/glad/generator/nim/templates/types/vk.nim
+++ b/glad/generator/nim/templates/types/vk.nim
@@ -1,0 +1,132 @@
+{.experimental: "codeReordering".}
+
+
+type
+# types required for: xcb
+  xcb_connection_t* = pointer
+  xcb_window_t* = uint32
+  xcb_visualid_t* = uint32
+# types required for: xlib(_xrandr)
+  Display* = pointer
+  RROutput* = uint
+  Window* = uint
+  VisualID* = uint
+# types required for: win32
+  BOOL* = cint
+  DWORD* = uint
+  LPVOID* = pointer
+  HANDLE* = pointer
+  HMONITOR* = pointer
+  WCHAR* = uint16
+  LPCWSTR* = ptr uint16
+  HINSTANCE* = pointer
+  HWND* = pointer
+  SECURITY_ATTRIBUTES* = object
+    nLength: DWORD
+    lpSecurityDescriptor: LPVOID
+    bInheritHandle: BOOL
+# types required for: wayland
+  wl_display* = pointer
+  wl_surface* = pointer
+# types required for: mir
+  MirConnection* = pointer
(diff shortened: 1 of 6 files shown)


glad/generator/nim/templates/types/vk.nim
```diff
@@ -0,0 +1,156 @@
+{.experimental: "codeReordering".}
```
krux02, line 1: as a former compiler developer, I recommend you to try to avoid this statement. I also don't think a simple binding should depend on any experimental features.

jacereda: Hey, thanks for the hint. I can try to remove it once the API is stable. I'm curious though, what kind of "surprises" should I expect from that feature?

krux02: Well, it is a bit complicated and not obvious why codeReordering is not trivial in Nim. In programming languages like Go where the top level order of definitions doesn't matter, there is a pre pass on all top level definitions, so that when function bodies are processed, all these symbols are known to the compiler. For Nim this is structurally impossible to do, since top level definitions can be generated by templates and macros.  Code reordering is a prepass that just changes the order of top level definitions according to their usage. This means you can first use a function and then later load the definition. The compiler will reorder this internally. But this prepass cannot be precise, because in order to know if a function A uses a function B it needs to fully semcheck function A, because the usages of B can be generated by macros and templates. And this is something that cannot be done yet, because we are still in the compilation prepass.

Regarding the surprises, macros in Nim may have sideeffects. It is not recommended to have side effects, but still it is possible. But the order in which macros are evaluated depends on the semchecking phase, so this could mess up some macros.

So in other words, for very simple Nim, where you don't use macros, templates or recursion, `codeReordering` might actually work, and this is probably the case here. But if your code evolves, and becomes more complex, you will realize more and more the shortcomings of code reordering and code reordering might become the reason why you don't understand the error messages anymore.

So I think it is bettor to just never get used to code reordering and get used to write code in the order the compiler expects it.

Not using code reordering also has an interesting side effect for the human reader, because it enforces an ordering of the code of a program. This helps a human to navigate a foreign codebase.

Dav1dde:
Thanks for your work, currently busy and the question I formulated in https://github.com/Dav1dde/glad/issues/302 kept me busy a bit, input welcome especially from contributors.

I will go over it when I get a minute.

Dav1dde:
The best way to move forward would be to develop this as a plugin. I cannot maintain all these languages with all the support they deserve. I will happily include plugins on the websites.

PR #300, opened 2020-09-27
Use GLAD_DEBUG to choose implementation at build time

nickgravelyn:
I'm not sure if this is something others are interested in, but I recently started using the C debug generator so I could put error checking after all my calls and I noticed that for a release build I'd have to re-gen the GLAD code since the debug version always includes the hooks.

So I made some changes for me so that instead of doing `#define GLAD_DEBUG` both the header and the implementation use `#if GLAD_DEBUG` to wrap up all the debug functionality such that I can define that for my debug builds and get my hooks while skipping it for release builds and automatically getting rid of the hooks entirely.

So for example in the header where the debug generator previously wrote:

```cpp
GLAPI PFNGLREADPIXELSPROC glad_glReadPixels;
GLAPI PFNGLREADPIXELSPROC glad_debug_glReadPixels;
#define glReadPixels glad_debug_glReadPixels
```

It would now write this:

```cpp
GLAPI PFNGLREADPIXELSPROC glad_glReadPixels;
#if GLAD_DEBUG
GLAPI PFNGLREADPIXELSPROC glad_debug_glReadPixels;
#define glReadPixels glad_debug_glReadPixels
#else
#define glReadPixels glad_glReadPixels
#endif
```

Which means my `glReadPixels` automatically goes straight to the non-debug version in non-debug builds. I also put the `#if` checks around the callback declarations, default implementations, and the glad_debug_* implementations just to ensure all that code is omitted in my release builds.

I haven't extensively tested if I've made some weird mistake but this seems to be working just fine for me building in Xcode on macOS so if it's useful in any way I'm happy to contribute it. If not, I'm happy to just keep this in our fork.

Thanks for the great library!

### DIFF
--- a/glad/lang/c/debug.py
+++ b/glad/lang/c/debug.py
@@ -11,14 +11,16 @@ DEFAULT_DEBUG_IMPL = '''
 
 
 DEBUG_HEADER = '''
-#define GLAD_DEBUG
+#if GLAD_DEBUG
 typedef void (* GLADcallback)(const char *name, void *funcptr, int len_args, ...);
 
 GLAPI void glad_set_pre_callback(GLADcallback cb);
 GLAPI void glad_set_post_callback(GLADcallback cb);
+#endif
 '''
 
 DEBUG_CODE = '''
+#if GLAD_DEBUG
 static GLADcallback _pre_call_callback = _pre_call_callback_default;
 void glad_set_pre_callback(GLADcallback cb) {
     _pre_call_callback = cb;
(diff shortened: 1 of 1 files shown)


Dav1dde:
[glad2](https://gen.glad.sh/) has install and uninstall functions at runtime to enable and disable the debugging functions:

```c
GLAD_API_CALL void gladInstallGLDebug(void);
GLAD_API_CALL void gladUninstallGLDebug(void);
```

Would this solve your problem?

nickgravelyn:
Oh yes that actually looks like it'll do exactly what I want. I guess I'll move over to glad2 then. Thanks!

PR #303, opened 2020-10-08
Fix for CMAKE_STAGING_PREFIX

jwinarske:
CMAKE_STAGING_PREFIX is used to stage artifacts prior to actual install step.  Then other build stages can use this "staged sysroot" using just a prefix.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -91,9 +91,7 @@ add_library(glad ${GLAD_SOURCES})
 add_dependencies(glad glad-generate-files)
 target_include_directories(glad
     PUBLIC
-        $<BUILD_INTERFACE:${GLAD_INCLUDE_DIRS}>
-    INTERFACE
-        $<INSTALL_INTERFACE:include/glad>)
+        $<BUILD_INTERFACE:${GLAD_INCLUDE_DIRS}>)
 
 # Are we building a shared library?
 get_target_property(library_type glad TYPE)
@@ -134,7 +132,7 @@ endif()
 # Install
 if(GLAD_INSTALL)
 
-  set(config_install_dir "lib/cmake/glad")
+  set(config_install_dir lib/cmake/glad)
   set(version_config "${CMAKE_CURRENT_BINARY_DIR}/gladConfigVersion.cmake")
   set(project_config "${CMAKE_CURRENT_BINARY_DIR}/gladConfig.cmake")
   set(targets_export_name "gladTargets")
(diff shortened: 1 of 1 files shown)


jwinarske:
Still needs some work.  I'm seeing /use/local/include/include/glad/glad.h

jwinarske:
It's ready for review

Dav1dde:
I am not very familiar with cmake, do you have an example on why this is needed?
Also is this a breaking change, it looks like it changes the include directory?

CMakeLists.txt
```diff
@@ -91,9 +91,7 @@ add_library(glad ${GLAD_SOURCES})
 add_dependencies(glad glad-generate-files)
 target_include_directories(glad
     PUBLIC
-        $<BUILD_INTERFACE:${GLAD_INCLUDE_DIRS}>
-    INTERFACE
-        $<INSTALL_INTERFACE:include/glad>)
```
Dav1dde, line 96: Why can this be removed?

CMakeLists.txt
```diff
@@ -158,26 +156,25 @@ if(GLAD_INSTALL)
   install(
     TARGETS glad
     EXPORT "${targets_export_name}"
-    LIBRARY DESTINATION "lib"
-    ARCHIVE DESTINATION "lib"
-    RUNTIME DESTINATION "bin"
-    INCLUDES DESTINATION "include")
-
-  if(GLAD_INCLUDE_DIRS)
```
Dav1dde, line 166: This check should still be necessary for the other languages?

CMakeLists.txt
jwinarske: This is one option, but not required.  Normally if you have a common header folder you can use the PUBLIC_HEADER property, and then include PUBLIC_HEADER in the INSTALL declaration.  Given the headers go into two different folders this isn't an option.  The "install(FILES" explicitly selects the files.

CMakeLists.txt
jwinarske: This should be language agnostic

jwinarske:
The premise of this change makes CMAKE_STAGING_PREFIX work as designed.

Say a cross-compiling project needs to build dependencies.  CMAKE_INSTALL_PREFIX sets the target install path.  CMAKE_STAGING_PREFIX sets the path for staging the cross-compiled binaries.  One usage pattern:

```
set(EXT_CMAKE_STAGING_PREFIX ${CMAKE_BINARY_DIR}/staging${CMAKE_INSTALL_PREFIX})

ExternalProject_Add(mapbox-gl-native
    GIT_REPOSITORY https://github.com/mapbox/mapbox-gl-native.git
    GIT_TAG master
    GIT_SHALLOW true
    SOURCE_DIR ${THIRD_PARTY_DIR}/mapbox-gl-native
    PATCH_COMMAND 
      git checkout . &&
      git apply ${CMAKE_SOURCE_DIR}/cmake/files/mapbox.patch
    BUILD_IN_SOURCE 0
    CMAKE_ARGS
        -D CMAKE_TOOLCHAIN_FILE=${CMAKE_BINARY_DIR}/target.toolchain.cmake
        -D CMAKE_PREFIX_PATH=${CMAKE_PREFIX_PATH}
        -D CMAKE_INSTALL_PREFIX=${CMAKE_INSTALL_PREFIX}
        -D CMAKE_STAGING_PREFIX=${EXT_CMAKE_STAGING_PREFIX}

```

In the case of generating executables CMAKE_INSTALL_PREFIX directly affects RPATH, which needs to match that of install path (target).  So CMAKE_STAGING_PREFIX is a critical variable.

Dav1dde:
Thanks for the explanations and the enhancement!

CMakeLists.txt
SpaceIm: I'm pretty sure it must not be removed, it's an information for the generated config files to set the include property on an installed package. But it was wrong, it should have been ` $<INSTALL_INTERFACE:include>`

Issue #314, opened 2021-03-01
Invalid source names generated when using CMake.


Aschratt:
The current C/C++ generator [appends a suffix](https://github.com/Dav1dde/glad/blob/master/glad/lang/c/generator.py#L36), if the selected spec does not match "gl". This causes problems when the project is used with CMake, since the *CMakeLists.txt* does [only add *glad.c*](https://github.com/Dav1dde/glad/blob/master/CMakeLists.txt#L77) to the sources.

To be more precise on what I am trying to do: I want to update the portfile for vcpkg to make it use features to select the spec. Everything works fine, but the build fails when a different spec than "gl" is specified. This happens, because the build cannot find the proper *glad.c* file.

I suggest to either drop the suffix from the generator (or make it optional), or alternatively select the sources based on the selected specification. I think the second option would be a better solution, however, I do not know how it would interfer with PR #276. 

There's also issue #244, which I think addresses a similar (or the same?) issue. It is quite unspecific though, so I'm not quite sure.

A possible fix could look something like this:

```cmake
if("${GLAD_SPEC}" STREQUAL "gl")
  list(APPEND GLAD_SOURCES "${GLAD_OUT_DIR}/src/glad.c")
else()
  list(APPEND GLAD_SOURCES "${GLAD_OUT_DIR}/src/glad_${GLAD_SPEC}.c")
endif()

list(APPEND GLAD_SOURCES "${GLAD_INCLUDE_DIRS}/glad/glad.h")
```

I do not want to open a PR, since I may also be wrong, so please correct me, if I missed something. 🙂

Dav1dde:
Fixed with #316

PR #316, opened 2021-03-10
Enhanced CMake support.

Aschratt:
This PR enhances the ability to generate GLAD (1) from CMake, but also addresses other generator issues. 

First of all, it fixes a bug that produces invalid builds when a different `GLAD_SPEC` than `gl` was specified (see issue #314). Since all other specs depend on the `gl` spec (i.e. the generated `glad.h` and `glad.c` code) to be present, I also had to introduce support for generating code for multiple specs, similar to PR #276. However, different to it, this PR does not introduce an additional command line argument, but instead extents `GLAD_SPEC` and `--spec` to support a list of specs instead. Furthermore, this PR also adds the generated headers to the install list, allowing to properly install the CMake target.

Since the `gl` spec is always required, it is added by default (if not present) within the CMakeLists.txt. It can be omitted in the `--spec` argument of the python generator scripts, in which case they produce the same output as before. The generator scripts now split the `--spec` argument and iterate all available scripts, running a generator on each. This, however, meant that I had to drop early validation for the argument (argparse `choices` did not work with `nargs` here 😕). If an unsupported spec is specified, no generator will be detected and the script fails.

When specifying multiple specs in combination with a set of extensions, the generator will fail, if it does not know the extension. However, there is no actual requirement for raising an error in this case, since it can also simply ignore the extension. I therefor removed the error from the generator and did not encounter any issues. This does not have any effect, if `GLAD_ALL_EXTENSIONS` is specified (or `--extension` is empty).

Additionally, I've improved handling for `wgl` and `egl` specs: 
- When generating code for the `wgl` spec, the generated code uses functions that are only defined, if `WGL_ARB_extensions_string` and `WGL_EXT_extensions_string`, hence the CMakeLists.txt adds those extensions, if it detects the `wgl` spec and does not find those extensions. This change does not effect the python generator script and does not have any effect, if `GLAD_ALL_EXTENSIONS` is enabled.
- The `egl` spec depends on the `eglplatform.h` header to be present. I've improved the generator to automatically download it from the egl registry, similar to how the `khrplatform.h` header is handled. It is only downloaded, if the `egl` spec is enabled and the `omit_khrplatform` flag is not set.

Furthermore, it should be possible to use GLAD as a git submodule with those changes and integrate it using CMake.

#### Things I could not test:

- Generating glx spec.
- Non C/C++ generators. Note that I did not do any obvious changes to the logic, so the generators should still work the same as previously.

#### Usage examples:

```shell
# Default build: generates default glad.h/glad.c ( `gl` spec) headers with all extensions enabled.
cmake -B ./build/
cmake --build ./build/

# Default build: generates default glad.h/glad.c ( `gl` spec) headers with no extensions.
cmake -DGLAD_ALL_EXTENSIONS=OFF -B ./build-min/
cmake --build ./build-min/

# Generate `gl`, `wgl` and `egl` specs
cmake -DGLAD_SPEC="wgl,egl" -B ./build-wgl-egl/
cmake --build ./build-wgl-egl/

# Same as above, but with installer.
cmake -DGLAD_SPEC="wgl,egl" -DGLAD_INSTALL=ON -DCMAKE_INSTALL_PREFIX="./install/" -B ./build-install/
cmake --build ./build-install/ --target install
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,9 @@ cmake_policy(SET CMP0048 NEW)
 # Enable MACOSX_RPATH by default
 cmake_policy(SET CMP0042 NEW)
 
+# Allow "IN_LIST" in "IF()"
+cmake_policy(SET CMP0057 NEW)
+
 if(NOT CMAKE_VERSION VERSION_LESS 3.1)
   # Silence warning about if()
   cmake_policy(SET CMP0054 NEW)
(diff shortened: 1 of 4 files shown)


Dav1dde:
@Aschratt I added a commit c4d86c4 to your fork, that should fix the problem from my last comment

Dav1dde:
Seems to be working now:

```
$ cmake -DGLAD_SPEC="gl,egl" -DGLAD_API="gles2=3.0" ..
```

If you don't have any concerns because of my changes or backwards compatibility, I would say this is ready.

Issue #319, opened 2021-04-18
[BUG] CMake Support on glad2


WildRackoon:
There are several issues with the glad2 cmake target:

- If `KEEP_SOURCES` if not set, do not expect the library to be linkable (with CMake `target_link_libraries`), as `gl.h` within `GLAD_DIR` would be deleted.
Actually, I dont see why glad sources should ever need to be deleted after generation at all ? (same for args.txt) its already cleaned when regeneration is needed.

- Building a SHARED glad2 lib does not work : 
`GLAD_API_CALL_EXPORT_BUILD` needs to be defined with `target_compile_definitions("${TARGET}" PUBLIC GLAD_API_CALL_EXPORT)` if `BUILD_SHARED_LIBS` is `ON`
Adding CMake's native handling of shared libs with `#if defined(GLAD_API_CALL_EXPORT_BUILD) || defined(glad_EXPORTS)` works great. So it would require some tweaking with `configure_file()` to keep the nice multi-glad/custom naming to insert ${TARGET} instead of my 'glad'.

- I understand the concern with `regenerate_${TARGET}`, but the `sleep and touch CMakeLists.txt` part is really sketchy and not proper cmake use: CMake will rerun the custom_command/custom_target it the `COMMAND` changes, so you just need to change the arguments and everything is fine (See https://gitlab.kitware.com/cmake/cmake/-/issues/20828).

- Glad sources should not be in the `CMAKE_BINARY_DIR`, but rather in glad2's `PROJECT_BINARY_DIR`
- args.txt could be configured into the code, and we could then define a `std::string glad_getConfigString()` ot retrieve it programmatically.
- The WriteFile.cmake is not necessary since it should be done at config time, no need to do it post-build
- Why isn't the `CMakeLists.txt` in the root of the repo ?

How is glad2 coming up so far ? If it is not in a stale state, I would be happy to give a try to rework the `CMakeLists.txt` if you want ?

#320

Dav1dde:
My cmake knowledge is very limited, the cmake setup itself has been contributed by others, for glad1 as well as glad2. All I do is try to keep it compatible and (somewhat) working.

I am thankful for every contribution that improves the cmake setup, as long as it remains compatible or there is a good reason for breaking something.

* Personally I am a fan of the glad2 setup using `glad_add_library`.
* `glad_getConfigString`: I don't want this in the API, I don't mind including it in an easily parseable comment though.
[...]

PR #320, opened 2021-04-19
Glad2 CMake Rework

WildRackoon:
Changes:
- Switched over `add_custom_command()` to create a file-based dependency with `OUTPUT ${GLAD_FILES} ${GLAD_ARGS_PATH}` on which added libraries depend on.
This `${TARGET}-generate` step is properly re-triggered only if one of its arguments changes and regeneration is needed.
Kept the `remove_directory()` for safety, but It may not be needed given how `GLAD_FILES` are very selective.
Sources should not be deleted afterwards, no reason to hide what we just generated, that's harder to debug.

- Fixed Shared Library building

- Not outrageous CMake minimal version bump to 3.2

- Allowed Interface Libraries for those with CMake 3.19+

- Removed unused `LIBTYPE`

Related to #319

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -6,7 +6,7 @@
 #    GLAD_SOURCES_DIR: path to the sources of glad (=python module)
 #
 #
-#  glad_add_library(<TARGET> [SHARED|STATIC|MODULE] [EXCLUDE_FROM_ALL] [MERGE] [QUIET] [KEEP_SOURCES] [LOCATION <PATH>]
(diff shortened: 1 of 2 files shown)


WildRackoon:
With regards to the CMakeLists.txt not being in ROOT, there is only a minor drawback to the most rigid among us:
CMake is meant to have one CMakeLists.txt per directory, with this setup an `add_subdirectory(glad2_src_dir/cmake)` is required and it will clutter the buildtree a bit with distinct `glad2_src_dir` for cmake files and our `gladsources`, more generally people prefer having everything in the same dir.

I talk again about this, because we could output the created libraries within their respective `gladsources/${TARGET}`, using:
```
set_target_properties( "${TARGET}"
    PROPERTIES
    LIBRARY_OUTPUT_DIRECTORY "${GLAD_DIR}"
)
```

PR #324, opened 2021-05-03
Require setuptools, glad/files/__init__.py imports pkg_resources

hrnciar:


Dav1dde:
This is an optional dependency, I think a better way would be to (optionally) switch to https://docs.python.org/3/library/importlib.html#module-importlib.resources

Instead of depending on setuptools, when it is not required for glad to work, similarly glad does not depend on lxml even though there is integration for it.

PR #326, opened 2021-05-11
Only write debug callback for GL spec.

### LINKED ISSUE #322: Ambiguous callback symbols when using non-gl debug loaders.
Say I want to load GL and WGL extensions into my application. I do this with something like this:

```cxx
auto hRC = ::wglCreateContext(hDC);
	
::wglMakeCurrent(hDC, hRC);
::gladLoadGL();
::gladLoadWGL(hDC);
```

From my understanding, this requires both loaders, i.e. `glad.c` and `glad_wgl.c` to be generated. 

However, in debug-mode, both source files contain the same definitions for `_pre_call_callback_default`/`_post_call_callback_default` and `glad_set_pre_callback`/`glad_set_post_callback`, causing errors when linking the code. 

Am I missing something or using the code in a wrong way or is this a bug?

I noticed, that the [this line](https://github.com/Dav1dde/glad/blob/7ece538856bf124d798ab323c8e1e64ebb83cb50/glad/lang/c/debug.py#L71) already writes different implementations, so an easy fix (imo) would be to prefix the function names.

crud89:
This fixes #322 by not outputting the callbacks to non-gl spec sources. This way, when a [glx,egl,wgl] source file is generated, it calls the `_pre_call_callback` and `_post_call_callback` functions defined in the gl source.

This slightly changes the behavior of spec callbacks, since `glad_glGetError` is now always evaluated in the `_post_call_callback` function. Earlier, spec sources simply ignored errors. If this was intentional, I will change the behavior to how it was before again and simply prefix the callback code, depending on the spec.

### DIFF
--- a/glad/lang/c/debug.py
+++ b/glad/lang/c/debug.py
@@ -1,55 +1,31 @@
 from glad.lang.c.generator import CGenerator
 
-
 DEFAULT_DEBUG_IMPL = '''
     {return_def}
-    _pre_call_callback("{name}", {args_callback});
+    _pre_call_callback{api}("{name}", {args_callback});
     {return_assign} glad_{name}({args});
-    _post_call_callback("{name}", {args_callback});
+    _post_call_callback{api}("{name}", {args_callback});
     {return_return}
 '''
 
-
 DEBUG_HEADER = '''
 #define GLAD_DEBUG
+#ifndef GLAD_CALLBACK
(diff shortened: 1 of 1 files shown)


Dav1dde:
I think the better way would be to rename the callbacks and include the API in the name, this way it still allows you to use a glx, egl or wgl loader without gl and get's rid of the `glGetError` call.

PR #330, opened 2021-06-10
C dbg output

Eugene-Kolesnikov:
For c-debug, added a registration of file name, line, and function name from which a gl command was called.


Dav1dde:
That's a great idea/feature, but not backwards compatible. Not sure how to handle this yet, I don't want to break compatibility.

Dav1dde:
I think due to compatibility concerns and little interest, I am closing this for now.

PR #332, opened 2021-07-15
c: zero-initializes parsed GL version numbers

okready:
If an unsupported GL_VERSION prefix string is encountered, the sscanf()
call used to parse the version number values may fail. When this occurs,
"major" and "minor" will be left uninitialized, which leads to undefined
behavior when using these values later.

This change addresses the issue by making sure "major" and "minor" are
initialized to zero. If the sscanf() call fails, subsequent checks
against these version number values will at least have consistent
results.


Dav1dde:
Thanks, good catch!

PR #333, opened 2021-07-15
glad: adds OpenGL SC support for GL_VERSION string parsing

okready:
OpenGL SC drivers return GL_VERSION strings with the format "OpenGL SC
N.M vendor-specific information", so the string "OpenGL SC " should be
included in the list of prefixes checked when parsing the GL_VERSION
string.


Dav1dde:
Thanks!

PR #341, opened 2021-10-27
nim: various improvements & fixes

johnnovak:
- use correct function call conventions (fixes crashes when compiling for 32-bit on Windows)
- differentiate between bitfields and enums
- general cleanup


Dav1dde:
Thanks as always! Sorry took me a bit.

PR #342, opened 2021-11-09
cmake: fix extensions parsing in PowerShell

Attractadore:
When EXTENSIONS is NONE, PowerShell passes a literary '' as an argument instead of an empty string. Use a string consisting of a single whitespace instead.


Attractadore:
Hi.
My use case is as follows:

glad_add_library(GL46 SHARED LANGUAGE C API gl:core=4.6 EXTENSIONS NONE)

This works fine on linux with bash, but when i try to build my project on windows using powershell (or cmd), I get an error:

  Generating with args --out-path D:/devel/KV-HW/triangle-intersection/build/src/Graphics/OpenGL/gladsources/GL46 --api gl:core=4.6 --extensions '' c
  [09.11.2021 18:07:17][INFO    ][glad          ]: getting 'gl' specification from remote location
  [09.11.2021 18:07:17][INFO        ][glad.opener        ]: opening: 'https://raw.githubusercontent.com/KhronosGroup/OpenGL-Registry/master/xml/gl.xml'
  invalid extensions or extensions not present in one of the selected APIs: ''

Using " " instead of "''" works on bash, powershell and cmd. I haven't tried any other shells.

Dav1dde:
Thanks! I am glad that works on Linux, seems like it's not just building a string and sending it through a shell.

PR #360, opened 2022-03-18
glad: add eglplatform.h for reproducible builds that includes egl

gregorburger:
eglplatform.h was missing for reproducible builds that include egl. Can be verified by:

`cmake -G Ninja -DGLAD_REPRODUCIBLE=ON -DGLAD_SPEC="egl" -DGLAD_API="gl=,gles2=" -DGLAD_ALL_EXTENSIONS=ON`


Dav1dde:
Thanks!

Issue #368, opened 2022-06-04
Hare generator


vladh:
Hello,

I've done work on extending glad to add a generator for the [Hare programming language](https://harelang.org/). It's currently in relatively good shape, and the OpenGL support in particular is used in some projects. However, there are still some issues to sort out, such as a problem that causes Vulkan extensions to not work correctly.

I wanted to ask what level of polish you would like to see in a generator before allowing it to be merged upstream. Would you like it to have full functionality and be polished? Or would you welcome a PR in a functional state that still requires some improvements, with subsequent PRs to add improvements?

Thanks!

Dav1dde:
Awesome that you're working on it! Are you making it for/with the glad 2 branch, this should make your life a lot easier.

Generally I do not want to accept new languages, that was one of the mistakes I made early, they only get second hand updates (when their initial maintainer is still around, luckily the ones I added so far have active maintainers that check in on issues!), but it still happens that they break every now and then (with new language versions or just a new specification that breaks something).

If you're developing against the glad 2 branch, you can make [a plugin](https://github.com/Dav1dde/glad/wiki/Extending-Glad), publish it on git(hub) and I can pin a version of that plugin for the web generator, I think this is the best way to go atm. Let me know if that works for you.

Dav1dde:
My current stance for this is, I cannot maintain other more languages with the support they deserve, I will help where I can and I intend to integrate mature plugins into the websites (and possibly as a git submodule).

PR #369, opened 2022-06-10
Implement is_descedant for Type and parse Type's parent and Member's enum

pipehuffer:
Having a way to filter Types by their parents allows Vulkan generators to load functions by type (similar to volk), and Member's enums can be used to set the array size of a particular Member.
This change is used with the generator mentioned in issue #368


Dav1dde:
Sorry I've set this sit for so long.

Issue #383, opened 2022-10-14
Don't load `_gl_handle` into static memory when using `--mx`


haasn:
Currently, `loader/gl.c` generates code like this when using it with `--mx` mode:

```c
static void* _gl_handle = NULL;

static void* glad_gl_dlopen_handle(void) {
    ....
    if (_gl_handle == NULL) {
        _gl_handle = glad_get_dlopen_handle(NAMES, sizeof(NAMES) / sizeof(NAMES[0]));
    }

    return _gl_handle;
}

int gladLoaderLoadGLContext(GladGLContext *context) {
    ...

    did_load = _gl_handle == NULL;
    handle = glad_gl_dlopen_handle();
    if (handle) {
        userptr = glad_gl_build_userptr(handle);

        version = gladLoadGLContextUserPtr(context,glad_gl_get_proc, &userptr);

        if (did_load) {
            gladLoaderUnloadGL();
        }
    }

    return version;
}
```

This is very racy (what if one thread calls `gladLoaderUnloadGL()` while another thread calls `gladLoaderLoadGLContext`?) and as far as I can tell, the `_gl_handle` variable serves no purpose here, because the only scope this handle is fully contained inside `gladLoaderLoadGLContext`, where it gets freed again afterwards.

As far as I can tell, the only purpose of this static variable is to be used in the case of `--on-demand`, in which case the handle needs to be persistent so functions can be loaded ad-hoc. So we should make its existence/use conditional on this flag.

I would go ahead and make this change, but I'm a bit confused as to why `gladLoaderUnloadGL` even exists in the case where `on-demand` is not used, because as far as I can tell, the non-ondemand loader is always designed to free the handle again after loading functions?

Dav1dde:
I generally never supported initialization from multiple threads for multiple reasons:

* most programs are single threaded (at least for the GL parts)
* multi threading with OpenGL is hard
* multi threading is hard, if I 'officially' support it, it needs to work forever, so I always went for simpler code or code that is easier on the templating
* preferred way to initialize would be with a loader from your windowing system
* glad is almost entirely made of globals in its core
* initialization can easily be synchronized outside of glad easily (e.g. using a mutex or writing your own loader)

[...]

Issue #372, opened 2022-07-23
--merge doesn't merge loaders?


haasn:
Hey,

I'm using GL in a way that's largely agnostic between GL core and GLES, so I'm naturally using the `--merge` option. One thing that surprised me is that, despite GL and GLES functions sharing the same function table now, `gladLoaderLoadGLContext` and `gladLoaderLoadGLES2Context` are still considered separate functions.

Right now, it seems I have to choose the right function to call based on inspecting `GL_VERSION` myself - which seems highly redundant with what glad is already doing.

Wouldn't it make much more sense for `--merge` to create a merged loader that loads the right API function set (GL or GLES2) depending on the `GL_VERSION` string automatically?

Dav1dde:
Why do you inspect `GL_VERSION`? For GLES you have to request a a different context, so the idea is you call the initializer for the context you requested.

haasn:
I am writing a library that gets implicitly passed a GL context (bound to the current thread) by the user, so I need to figure out which version of the context to load.

While I *could* require the user specify up-front whether it's a Desktop or ES context, this would be a rather annoying non-backwards compatible change because previous versions of the library (based on libepoxy instead of glad) did not require this.

Dav1dde:
For glad GL and GLES are still separate APIs the `--merge` option only collapses the identically named symbols, but the whole loading machinery still loads separately. E.g. glad won't attempt to load a GL only function when instructed to load GLES, the same is true for symbols like `GLAD_GL_ES_VERSION_*` and `GLAD_GL_VERSION_*`.

So depending on what you load you get different symbols, which is properly represented in the API.

I currently don't think I want to introduce a "Auto Loader" function/facade which depending on the version returned calls the correct loader.

Depending on your usecase you can always call both loaders (Note: you wont be able to rely on `GLAD_*_VERSION_*` constants), or maybe the on-demand generation function is helpful (which is very close to libepoxy) and with on-demand you can set the same loader for both APIs.

Out of curiosity, how do you decide which loader to pass to glad, since this is also context dependent, or are you using GLFW/SDL which take care of this?

haasn:
I think for the time being I will just call both loader functions, in the hopes of picking up every function in the end. The risk of a "false positive" (i.e. trying to load a GLES function on a GL) is pretty negligible in practice, anyway. (And actually, sort of required - because in practice, I need to load some GLES functions even on Desktop GL contexts - it's complicated)

[...]

PR #384, opened 2022-10-14
[RFC] c: make mx context thread safe

haasn:
There is no reason to store the handle in `_gl{es}_handle` when not using an ondemand context, because all of the other context types can immediately unload the handle after using it, as part of the same scope.

This fixes a potential race condition when two threads try loading the handle at the same time.

Fixes: #383

### DIFF
--- a/glad/generator/c/templates/loader/egl.c
+++ b/glad/generator/c/templates/loader/egl.c
@@ -20,8 +20,6 @@ static GLADapiproc glad_egl_get_proc(void *vuserptr, const char* name) {
     return result;
 }
 
-static void* _egl_handle = NULL;
-
 static void* glad_egl_dlopen_handle(void) {
 #if GLAD_PLATFORM_APPLE
     static const char *NAMES[] = {"libEGL.dylib"};
@@ -31,11 +29,7 @@ static void* glad_egl_dlopen_handle(void) {
     static const char *NAMES[] = {"libEGL.so.1", "libEGL.so"};
 #endif
 
-    if (_egl_handle == NULL) {
-        _egl_handle = glad_get_dlopen_handle(NAMES, sizeof(NAMES) / sizeof(NAMES[0]));
-    }
-
-    return _egl_handle;
+    return glad_get_dlopen_handle(NAMES, sizeof(NAMES) / sizeof(NAMES[0]));
 }
 
 static struct _glad_egl_userptr glad_egl_build_userptr(void *handle) {
@@ -49,10 +43,8 @@ static struct _glad_egl_userptr glad_egl_build_userptr(void *handle) {
 int gladLoaderLoadEGL(EGLDisplay display) {
     int version = 0;
     void *handle = NULL;
-    int did_load = 0;
     struct _glad_egl_userptr userptr;
 
-    did_load = _egl_handle == NULL;
     handle = glad_egl_dlopen_handle();
     if (handle != NULL) {
         userptr = glad_egl_build_userptr(handle);
@@ -61,9 +53,7 @@ int gladLoaderLoadEGL(EGLDisplay display) {
             version = gladLoadEGLUserPtr(display, glad_egl_get_proc, &userptr);
         }
 
-        if (!version && did_load) {
-            gladLoaderUnloadEGL();
-        }
+        glad_close_dlopen_handle(handle);
     }
 
     return version;
@@ -82,13 +72,12 @@ static GLADapiproc glad_egl_internal_loader_get_proc(const char *name) {
 {% endif %}
 
 void gladLoaderUnloadEGL() {
-    if (_egl_handle != NULL) {
-        glad_close_dlopen_handle(_egl_handle);
-        _egl_handle = NULL;
 {% if options.on_demand %}
+    if (glad_egl_internal_loader_global_userptr.handle != NULL) {
+        glad_close_dlopen_handle(glad_egl_internal_loader_global_userptr.handle);
         glad_egl_internal_loader_global_userptr.handle = NULL;
-{% endif %}
     }
+{% endif %}
 }
 
 #endif /* GLAD_EGL */
--- a/glad/generator/c/templates/loader/gl.c
+++ b/glad/generator/c/templates/loader/gl.c
@@ -23,8 +23,6 @@ static GLADapiproc glad_gl_get_proc(void *vuserptr, const char *name) {
     return result;
 }
 
-static void* _gl_handle = NULL;
-
 static void* glad_gl_dlopen_handle(void) {
 #if GLAD_PLATFORM_APPLE
     static const char *NAMES[] = {
@@ -45,11 +43,7 @@ static void* glad_gl_dlopen_handle(void) {
     };
 #endif
 
-    if (_gl_handle == NULL) {
-        _gl_handle = glad_get_dlopen_handle(NAMES, sizeof(NAMES) / sizeof(NAMES[0]));
-    }
-
-    return _gl_handle;
+    return glad_get_dlopen_handle(NAMES, sizeof(NAMES) / sizeof(NAMES[0]));
 }
 
 static struct _glad_gl_userptr glad_gl_build_userptr(void *handle) {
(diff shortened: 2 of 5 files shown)


haasn:
Note: This commit currently only fixes `gl.c`, but of course, *all* of the loaders share the same issue. It would be nice to get some feedback on it before I go ahead and do the manual work of updating all of the other loaders.

haasn:
Also note that this commit makes the `Unload` family of functions pointless no-ops in the non-ondemand case, preserved merely for API backwards compatibility. But it might be clearer to also drop the API function in the case, to make it clearer to the user that unloading is not required when not using `--ondemand`.

haasn:
After further reflection on this code, I've instead opted to remove `_gl_handle` entirely.

haasn:
Implemented something similar for the rest of the handle types, with the notable exception of `glx`. Let me know what you think about this approach.

Dav1dde:
This does not work for any loader which needs symbols from the .dll/.so, like EGL. But it is even risky for GL itself (1.1 symbols on Windows or I am not sure how ANGLE works here). So I would keep it as it is and just store the user pointer for MX on the MX struct instead of globally.

haasn:
> So I would keep it as it is and just store the user pointer for MX on the MX struct instead of globally.

I looked into this way of doing it but I was stuck on the fact that the `ondemand` code-path still requires accessing the handle, which it can no longer if it's inside the MX context.

Also, what do you mean by "user pointer" here? Isn't this about the `void* handle`? How does the user pointer factor into it?

Dav1dde:
> I looked into this way of doing it but I was stuck on the fact that the `ondemand` code-path still requires accessing the handle, which it can no longer if it's inside the MX context.

That's fine we don't have to support this, on-demand + MX does not really make sense and is currently not supported: https://github.com/Dav1dde/glad/blob/68c595d307697a60749a41425e64a1e2120b5716/glad/generator/c/__init__.py#L243-L248

> Also, what do you mean by "user pointer" here? Isn't this about the `void* handle`? How does the user pointer factor into it?

Sorry worded that badly, I meant we should add a `void *userptr` to the generated MX struct and store the handle in it.

The `*userptr` should be usable by anyone, not just a hidden detail of the builtin loader. I feel it could be very useful in general if a user can associate data with a context (be it for their own loader or other data). It also should be the first field (so it's always at a fixed location in the struct).

The builtin loader can just use the `*userptr` field to store the dlsym handle.

haasn:
> Sorry worded that badly, I meant we should add a void *userptr to the generated MX struct and store the handle in it.

Sorry for asking maybe obvious questions, but just to clarify, this field already exists:

https://github.com/Dav1dde/glad/blob/glad2/glad/generator/c/templates/base_template.h#L68

And it is obviously where the userptr is already stored. But we can't just re-use this directly for the internal loader, because there's no obvious way to distinguish between user-provided userptr and internally loaded handles. And by the way, when using `--merge` it is possible to load e.g. GL and GLES2 functions to the same context, with two opened handles.

Maybe what we actually want to be doing is moving the `_gl_handle` etc. fields into the context struct? So next to the `void *usrptr` we have `void *_gl_handle, *_gles1_handle, ...`, one for each type of loader.

Dav1dde:
> Sorry for asking maybe obvious questions, but just to clarify, this field already exists

Huh, I guess past me was smarter than current me.

Is there really a usecase to load 2 APIs at the same time into the MX struct? It should be the user's responsibility to remember which loader was used to call the correct unload function again.

I agree it sucks if we take the userptr, first thing that comes to mind is to add an internal user pointer only if the loader option is enabled (this could just be a `#ifdef` maybe) at the end of the struct.

haasn:
> Is there really a usecase to load 2 APIs at the same time into the MX struct?

I need it for interop between `EGLImageKHR` (e.g. as imported by vaapi `dmabuf`) and GL textures. The only extension I can find to implement this functionality is `GL_OES_EGL_image_external`, an OpenGL ES extension. Turns out if you just load the functions it works fine, but you have to load `GLES2` pointers from the same context for this to work.

Dav1dde:
This is a very special use case and wanders in the areas of undefined behaviour and driver implementation specifics. I dont think this should be supported, especially with the builtin loader, which loads different drivers for the potential symbols. You can do this if you supply your own loader or patch the generated code.

[Kitty used to do this](https://github.com/kovidgoyal/kitty/blob/master/glad/generate.py), they have a script to re-generate the loader, with some post processing but the generated code is checked in (also what I recommend).

haasn:
@Dav1dde even with that use case ironed out, there remains the issue that you explicitly advised me to call into both loader functions in #372, as a resolution to glad2's inability to figure out the right GL context type (GL vs GLES) at runtime.

The problem is that right now, the only way to write code that works on both GL and GLES is to essentially write the following:

```c
GladGLContext gl = {0};
bool loaded = false;
loaded |= gladLoaderLoadGLContext(&gl);
loaded |= gladLoaderLoadGLES2Context(&gl);
if (loaded) {
    const char *version = gl.GetString(GL_VERSION);
    is_gles = strncmp(version, "OpenGL ES", strlen("OpenGL ES"));
}
```

Dav1dde:
> My preferred resolution to that, of course, would still be to have --merge also merge loaders

This would be broken behavior and leads to more problems down the road.

> as a resolution to glad2's inability to figure out the right GL context type (GL vs GLES) at runtime

This is not an inability, this is the job of the caller.

In the linked issue I did not advise you to do it, I said it is possible. You should realize that this puts glad into an undefined state, depending on how the drivers react. You seem to have some misconceptions here, like:

> I suppose one remaining downside is that this implies unnecessarily calling dlopen / dlclose twice, when once would suffice with a merged loader.

This would definitely not suffice, GL and GLES have entirely different drivers. The builtin loaders also `dlopen` different shared libraries. I don't even know if you're guaranteed a proper return value when calling GLES `glGetString` with a GL context active or if drivers are allowed to crash in this case.

Figuring out which specific context is currently active is not the job of glad, this is way beyond the scope. If it happens to work for your usecase, that's fine, it definitely is not an intended feature.

The correct way would be to figure out what context you're running under *before* calling any glad loader. You have so many issues here, one of the largest being, the internal loader can't even know the correct driver to load. E.g. it attempts to load `GLESv2.dll`, on Windows, just to get a handle to `eglGetProcAddress`, but this DLL does not need to be available (or be the *correct* driver at all), what if ANGLE is actually packaged in a non standard way, or there are multiple DLLs available. If you are not the application, but a library you cannot depend on *any* of these circumstances and might just be rewarded with a segfault.

Maybe all these things are actually specified or you are okay to rely on undefined behavior of driver implementations, I can't guarantee sound behavior here and that's why this is not supported (and probably won't be).

If you really want to ignore these issues, the fix is simple, you load into different structs and see which one succeeded, then discard the other struct. For unloading you would have to remember the context (`userptr`?) or just query the version once more.

Pseudo Code:

Load:
```c
GladGLContext gl = {0};
if (gladLoaderLoadGLContext(&gl)) {
   return gl;
}
GladGLContext gles2 = {0};
if (gladLoaderLoadGLES2Context(&gles2)) {
   return gles2;
}
```

Unload:
```c
if (gl.GetString(GL_VERSION) contains "GLES") {
    gladLoaderLoadUnloadGLES2Context(&gl);
} else {
    gladLoaderLoadUnloadGLContext(&gl);
}
```

haasn:
> The correct way would be to figure out what context you're running under before calling any glad loader.

The reason I'm so reluctant to do this is because it's an API break on my part as well, since under libepoxy it was not required to specify this. But maybe it's the only (sane) way after all. I'll think about it.

> If you really want to ignore these issues, the fix is simple, you load into different structs and see which one succeeded, then discard the other struct. For unloading you would have to remember the context (userptr?) or just query the version once more.

This approach doesn't work un-modified because `gladLoaderLoadGLContext` actually succeeds on a GLES context, unfortunately. So you'd need some extra checks to see if the loaded GL version contains `OpenGL ES`, and if so, unload GL and re-load as GLES. But it's possible in principle. (Or maybe `gladLoaderLoadGLContext` should *reject* contexts which have `OpenGL ES` in their version string? Is this a bug in glad2?)

But in any case, I can easily extend this approach in some way that works, I suppose. I guess the take-away at the end of the day here is that having a single `void *handle` inside the mx context would work, after fixing all of the above.

Dav1dde:
> The reason I'm so reluctant to do this is because it's an API break on my part as well, since under libepoxy it was not required to specify this. But maybe it's the only (sane) way after all. I'll think about it.

I get that and I am sure we can find a solution which works. I just don't to expose this potentially broken/undefined behavior exposed through a glad API.

> Or maybe gladLoaderLoadGLContext should reject contexts which have OpenGL ES in their version string? Is this a bug in glad2?

I think so.

But this shouldn't be an issue on your end apart from trying GLES before GL in my pseudo code.

Do you want to still work on changing the unload signature and storing the internal handle on the context?

Dav1dde:
> But this shouldn't be an issue on your end apart from trying GLES before GL in my pseudo code.

Thinking about this, I dont think that works. You'd have to check `glGetString` yourself.

haasn:
Superseded by #385 

> Thinking about this, I dont think that works. You'd have to check glGetString yourself.

I just wrote code for this. It's not too bad.

PR #385, opened 2022-10-20
c: store handles inside mx struct

haasn:
Altered approach, as discussed in #384, which this PR supersedes.

TODO:
- [x] gles1
- [x] vulkan
- [x] update docs?
- [x] update examples?
- [x] update tests?

Do you prefer this approach?

### DIFF
--- a/example/c++/hellowindow2_macro.cpp
+++ b/example/c++/hellowindow2_macro.cpp
@@ -64,7 +64,7 @@ int main()
 #endif
 
 #ifdef GLAD_OPTION_GL_MX
-    GladGLContext context;
+    GladGLContext context = {};
   #ifdef GLAD_OPTION_GL_LOADER
     int version = gladLoaderLoadGLContext(&context);
   #else
--- a/example/c++/hellowindow2_mx.cpp
+++ b/example/c++/hellowindow2_mx.cpp
@@ -48,7 +48,7 @@ int main()
     // Set the required callback functions
     glfwSetKeyCallback(window, key_callback);
 
-    GladGLContext context;
+    GladGLContext context = {};
     int version = gladLoadGLContext(&context, glfwGetProcAddress);
     if (version == 0)
     {
--- a/glad/generator/c/templates/base_template.h
+++ b/glad/generator/c/templates/base_template.h
@@ -76,6 +76,10 @@ typedef struct Glad{{ feature_set.name|api }}Context {
     {{ command.name|pfn }} {{ command.name|ctx(member=True) }};
 {% endcall %}
 {% endfor %}
+
+{% if options.loader %}
+    void* glad_loader_handle;
+{% endif %}
 } Glad{{ feature_set.name|api }}Context;
 
 {% if options.mx_global %}
--- a/glad/generator/c/templates/loader/gl.c
+++ b/glad/generator/c/templates/loader/gl.c
@@ -23,9 +23,11 @@ static GLADapiproc glad_gl_get_proc(void *vuserptr, const char *name) {
     return result;
 }
 
-static void* _gl_handle = NULL;
+{% if not options.mx %}
+static void* {{ template_utils.handle() }} = NULL;
+{% endif %}
 
-static void* glad_gl_dlopen_handle(void) {
+static void* glad_gl_dlopen_handle({{ template_utils.context_arg(def='void') }}) {
 #if GLAD_PLATFORM_APPLE
     static const char *NAMES[] = {
         "../Frameworks/OpenGL.framework/OpenGL",
@@ -45,11 +47,11 @@ static void* glad_gl_dlopen_handle(void) {
     };
 #endif
 
-    if (_gl_handle == NULL) {
-        _gl_handle = glad_get_dlopen_handle(NAMES, sizeof(NAMES) / sizeof(NAMES[0]));
+    if ({{ template_utils.handle() }} == NULL) {
+        {{ template_utils.handle() }} = glad_get_dlopen_handle(NAMES, sizeof(NAMES) / sizeof(NAMES[0]));
     }
 
-    return _gl_handle;
+    return {{ template_utils.handle() }};
 }
 
 static struct _glad_gl_userptr glad_gl_build_userptr(void *handle) {
@@ -76,15 +78,15 @@ int gladLoaderLoadGL{{ 'Context' if options.mx }}({{ template_utils.context_arg(
     int did_load = 0;
     struct _glad_gl_userptr userptr;
 
-    did_load = _gl_handle == NULL;
-    handle = glad_gl_dlopen_handle();
+    did_load = {{ template_utils.handle() }} == NULL;
+    handle = glad_gl_dlopen_handle({{ 'context' if options.mx }});
     if (handle) {
         userptr = glad_gl_build_userptr(handle);
 
         version = gladLoadGL{{ 'Context' if options.mx }}UserPtr({{ 'context,' if options.mx }}glad_gl_get_proc, &userptr);
 
         if (did_load) {
-            gladLoaderUnloadGL();
+            gladLoaderUnloadGL{{ 'Context' if options.mx }}({{ 'context' if options.mx }});
         }
     }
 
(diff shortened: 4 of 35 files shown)


glad/generator/c/templates/base_template.h
```diff
@@ -66,6 +66,9 @@ extern "C" {
 {% else %}
 typedef struct Glad{{ feature_set.name|api }}Context {
     void* userptr;
+{% if options.loader %}
+    void* handle;
```
Dav1dde, line 70: This should be at the end of the struct and be named `glad_loader_handle` to make sure this doesn't 'accidentally' get used for anything else by a user.

haasn: Fixed, and also added the `_glad` prefix to the statically generated variable for consistency. (Now that we can easily use a longer name due to the template macro)

Dav1dde:
> As a side note, how come EGL isn't affected by `--mx`?

Correct, it is only implemented for context specific APIs and ignored for all others (EGL, WGL, GLX). It's not an error to allow generating multiple APIs at once.

haasn:
I believe the reason the test fails is because it leaves `GladGLContext` uninitialized when calling into `gladLoaderLoadGLContext`, which has the result of triggering the `if (handle) { return handle; } ` logic inside the loader.

I'm not sure whether this is supposed to be supported behavior or not, but to be on the safe side, I think we should memset to 0 either the entire struct, or at least the handle, inside `gladLoaderLoadContext()`. Thoughts?

Dav1dde:
> I'm not sure whether this is supposed to be supported behavior or not, but to be on the safe side, I think we should memset to 0 either the entire struct, or at least the handle, inside `gladLoaderLoadContext()`. Thoughts?

Yeah the context should be zero initialized, this is wrong in the current tests.

Dav1dde:
The examples also would need that update, [this](https://github.com/Dav1dde/glad/blob/glad2/example/c%2B%2B/multiwin_mx/multiwin_mx.cpp#L81) currently uses `malloc` this should be `calloc`.

haasn:
Updated the failing gl tests, also the examples, and all of the compile-only tests.

Dav1dde:
Now the last test is failing because the global instance is not a pointer, but it should be. But this also requires more changes, I didn't want to make, but now there is a good reason to.

I'll fix the mx global stuff tomorrow, then all tests should pass.

haasn:
I think the `--mx-global` test is failing because of the way the `c` loader unloads the handle again even after a successful load (a change which does not get reflected back to the global context, thus triggering the assert). This is decidedly different from all the other loaders. Is this a bug?

haasn:
Even with the above fixed, the following is now undefined behavior:

```c
// build with --mx --mx-global
GladGLContext context = {0};
gladLoaderLoadGLContext(&context);
gladLoaderUnloadGLContext(&context);
gladLoaderUnloadGLContext(gladGetGLContext());
```

We could make this defined behavior by having `gladLoaderUnloadGLContext()` also clear the global context when using `--mx-global`. (To be honest, I'm not sure why `--mx-global` even exists - it seems to be defeating its own reason to exist, no?)

Thoughts?

Dav1dde:
The problem is, the globally stored context is not a pointer, when setting the context it's dereferenced/copied. The global context should be a pointer with a default instance in static memory.

Dav1dde:
> We could make this defined behavior by having gladLoaderUnloadGLContext() also clear the global context when using --mx-global. 

This is fine after I change the global instance to a pointer (what it should've been), since the attached handle will be cleared correctly.

> (To be honest, I'm not sure why --mx-global even exists - it seems to be defeating its own reason to exist, no?)

It exists as a compatibility layer for older code, you can switch the context and you dont have to update code. Unfortunately I dont have actual usage stats, because this might very well be unused in practice.

haasn:
@Dav1dde like this?

```diff
diff --git a/glad/generator/c/templates/base_template.c b/glad/generator/c/templates/base_template.c
index 640d183..df04d19 100644
--- a/glad/generator/c/templates/base_template.c
+++ b/glad/generator/c/templates/base_template.c
@@ -20,7 +20,7 @@ extern "C" {
 
 {% block variables %}
 {% if options.mx_global %}
-{% call template_utils.zero_initialized() %}Glad{{ feature_set.name|api }}Context {{ global_context }}{% endcall %}
+Glad{{ feature_set.name|api }}Context* {{ global_context }};
 {% endif %}
 {% endblock %}
 
diff --git a/glad/generator/c/templates/base_template.h b/glad/generator/c/templates/base_template.h
index 4773636..4583ef7 100644
--- a/glad/generator/c/templates/base_template.h
+++ b/glad/generator/c/templates/base_template.h
@@ -83,14 +83,14 @@ typedef struct Glad{{ feature_set.name|api }}Context {
 } Glad{{ feature_set.name|api }}Context;
 
 {% if options.mx_global %}
-GLAD_API_CALL Glad{{ feature_set.name|api }}Context glad_{{ feature_set.name }}_context;
+GLAD_API_CALL Glad{{ feature_set.name|api }}Context* glad_{{ feature_set.name }}_context;
 
 {% for extension in chain(feature_set.features, feature_set.extensions) %}
-#define GLAD_{{ extension.name }} (glad_{{ feature_set.name }}_context.{{ extension.name|no_prefix }})
+#define GLAD_{{ extension.name }} (glad_{{ feature_set.name }}_context->{{ extension.name|no_prefix }})
 {% endfor %}
 
 {% for command in feature_set.commands %}
-#define {{ command.name }} (glad_{{ feature_set.name }}_context.{{ command.name|no_prefix }})
+#define {{ command.name }} (glad_{{ feature_set.name }}_context->{{ command.name|no_prefix }})
 {% endfor %}
 {% endif %}
 
diff --git a/glad/generator/c/templates/gl.c b/glad/generator/c/templates/gl.c
index 9cf9bce..218cfd6 100644
--- a/glad/generator/c/templates/gl.c
+++ b/glad/generator/c/templates/gl.c
@@ -230,11 +230,11 @@ int gladLoad{{ api|api }}(GLADloadfunc load) {
 
 {% if options.mx_global %}
 Glad{{ feature_set.name|api }}Context* gladGet{{ feature_set.name|api }}Context() {
-    return &{{ global_context }};
+    return {{ global_context }};
 }
 
 void gladSet{{ feature_set.name|api }}Context(Glad{{ feature_set.name|api }}Context *context) {
-    {{ global_context }} = *context;
+    {{ global_context }} = context;
 }
 {% endif %}
 ```

I don't think there needs to be a default instance - before an instance is loaded, attempting to use global functions will segfault one way or the other.

**Comment by Dav1dde** [comment]
Yeah the patch looks like what I had in mind.

> I don't think there needs to be a default instance - before an instance is loaded, attempting to use global functions will segfault one way or the other.

There is for the loader ([here](https://github.com/Dav1dde/glad/blob/glad2/glad/generator/c/templates/gl.c#L203-L205)):

```c
{% if options.mx_global %}
int gladLoad{{ api|api }}(GLADloadfunc load) {
    return gladLoad{{ api|api }}Context(gladGet{{ feature_set.name|api }}Context(), load);
}
{% endif %}
```

glad/generator/c/templates/loader/gl.c
```diff
@@ -76,15 +78,15 @@ int gladLoaderLoadGL{{ 'Context' if options.mx }}({{ template_utils.context_arg(
...
+    did_load = {{ template_utils.handle() }} == NULL;
+    handle = glad_gl_dlopen_handle({{ 'context' if options.mx }});
     if (handle) {
         userptr = glad_gl_build_userptr(handle);
 
         version = gladLoadGL{{ 'Context' if options.mx }}UserPtr({{ 'context,' if options.mx }}glad_gl_get_proc, &userptr);
 
-        if (did_load) {
-            gladLoaderUnloadGL();
+        if (!version && did_load) {
```
Dav1dde, line 88: This shouldn't be required anymore

example/c++/hellowindow2_macro.cpp
```diff
@@ -64,7 +64,7 @@ int main()
 #endif
 
 #ifdef GLAD_OPTION_GL_MX
-    GladGLContext context;
+    GladGLContext context = {0};
```
Dav1dde, line 67: This is a CPP file, it should be zero initialized with `{}`

example/c++/hellowindow2_mx.cpp
```diff
@@ -48,7 +48,7 @@ int main()
     // Set the required callback functions
     glfwSetKeyCallback(window, key_callback);
 
-    GladGLContext context;
+    GladGLContext context = {0};
```
Dav1dde, line 51: This is a CPP file, it should be zero initialized with `{}`

Dav1dde:
@haasn thanks for your work and time!

Dav1dde:
PS: I've seen you have `glad` as a git submodule, any particular reason you don't just include the generated files but integrate glad into your build?

haasn:
> PS: I've seen you have `glad` as a git submodule, any particular reason you don't just include the generated files but integrate glad into your build?

I think it's bad style to check generated files into version control. Also, I feel that integrating it into the build system directly makes it more transparent and reproducible, especially since I whitelist the extensions to generate.

But it is always an option to keep in the back of my mind, that's true.

PR #386, opened 2022-10-21
c: store mx-global context as a pointer

haasn:
Instead of a separate struct that needs to be copied by-value. This will reduce the risk of accidentally ending up with an inconsistent context state.

We still need a default context in the even that a user loads an `mx-global` context without ever supplying a custom struct, but this is easy enough to return in an ad-hoc fashion whenever `gladGet*Context` is called without a context otherwise being bound.

(Incidentally, the way this logic is implemented also lets you recover the global context by explicitly calling `gladSet*Context(NULL)`)

### DIFF
--- a/glad/generator/c/templates/base_template.c
+++ b/glad/generator/c/templates/base_template.c
@@ -20,7 +20,7 @@ extern "C" {
 
 {% block variables %}
 {% if options.mx_global %}
-{% call template_utils.zero_initialized() %}Glad{{ feature_set.name|api }}Context {{ global_context }}{% endcall %}
+Glad{{ feature_set.name|api }}Context* {{ global_context }};
 {% endif %}
 {% endblock %}
 
--- a/glad/generator/c/templates/base_template.c
+++ b/glad/generator/c/templates/base_template.c
@@ -20,7 +20,8 @@ extern "C" {
 
 {% block variables %}
 {% if options.mx_global %}
-Glad{{ feature_set.name|api }}Context* {{ global_context }};
+{% call template_utils.zero_initialized() %}Glad{{ feature_set.name|api }}Context {{ global_context }}_static{% endcall %}
+Glad{{ feature_set.name|api }}Context* {{ global_context }} = &{{ global_context }}_static;
 {% endif %}
 {% endblock %}
 
--- a/glad/generator/c/templates/base_template.h
+++ b/glad/generator/c/templates/base_template.h
@@ -79,14 +79,14 @@ typedef struct Glad{{ feature_set.name|api }}Context {
 } Glad{{ feature_set.name|api }}Context;
 
 {% if options.mx_global %}
-GLAD_API_CALL Glad{{ feature_set.name|api }}Context glad_{{ feature_set.name }}_context;
+GLAD_API_CALL Glad{{ feature_set.name|api }}Context* glad_{{ feature_set.name }}_context;
 
 {% for extension in chain(feature_set.features, feature_set.extensions) %}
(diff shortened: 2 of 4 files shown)


haasn:
Sorry for jumping the gun on this even though you said you'd write the patch yourself, but I am currently blocked by this (and the lack of a glad2 release) for my own work :p

haasn:
Huh? The failing test `test/c/compile/gl/mx+mx-global/008` works on my end.

Dav1dde:
@haasn I've moved the static handle into the global scope, what do you think?

If somebody uses the global handle, overrides it without previously saving it, it is gone, but that is fine in my eyes. I am also planning to disable MX_GLOBAL before 2.0, as you mentioned this is tricky and somewhat goes against the usefulness of MX. If there are actual users, I will consider enabling it again.

haasn:
> @haasn I've moved the static handle into the global scope, what do you think?
> 
> If somebody uses the global handle, overrides it without previously saving it, it is gone, but that is fine in my eyes. I am also planning to disable MX_GLOBAL before 2.0, as you mentioned this is tricky and somewhat goes against the usefulness of MX. If there are actual users, I will consider enabling it again.

Might be cleaner to do it this way, yeah. Though `gladGet*Context()` is not a hot-path, so I don't see why not do the check/defaulting there. And it does preserve the ability to recover the default context by using `gladSetGLContext(NULL)` if you keep it inside there. But I don't particularly care either way.

P.s. you don't need to zero-initialize static variables

haasn:
At least that seems to have fixed the CI issue. Welp. I was gonna try debugging with something like this:

```diff
diff --git a/test/c/compile/gl/mx+mx-global/008/test.c b/test/c/compile/gl/mx+mx-global/008/test.c
index a4c37c2..3125229 100644
--- a/test/c/compile/gl/mx+mx-global/008/test.c
+++ b/test/c/compile/gl/mx+mx-global/008/test.c
@@ -8,7 +8,7 @@
  *
  * GLAD: $GLAD --out-path=$tmp --api="gl:compatibility" c --loader --mx --mx-global
  * COMPILE: $GCC -c -o $tmp/test.o -I$tmp/include $tmp/src/gl.c -ldl
- * RUN: [ "$(nm $tmp/test.o | grep ' C ' | wc -l)" -eq "0" ]
+ * RUN: [ "$(nm $tmp/test.o | grep ' C ' | wc -l)" -eq "0" ] || nm $tmp/test.o
  */
 
 #error "this testfile should not be compiled"
diff --git a/utility/test.sh b/utility/test.sh
index 57161bd..be25132 100755
--- a/utility/test.sh
+++ b/utility/test.sh
@@ -9,7 +9,7 @@ GLAD=${GLAD:="$PYTHON -m glad ${GLAD_ARGS}"}
 
 _GCC=${_GCC:="gcc"}
 _MINGW_GCC=${_MINGW_GCC:="x86_64-w64-mingw32-gcc"}
-_GCC_FLAGS="-Wall -Wextra -Werror -Wsign-conversion -Wcast-qual -Wstrict-prototypes -ansi -pedantic"
+_GCC_FLAGS="-Wall -Wextra -Werror -Wsign-conversion -Wcast-qual -Wstrict-prototypes -ansi -pedantic -g"
 
 GCC=${GCC:="$_GCC $_GCC_FLAGS"}
 MINGW_GCC=${MINGW_GCC:="$_MINGW_GCC $_GCC_FLAGS"}
```

but it seems it's no longer needed. (Incidentally, putting `-g` into the test flags would be a good idea in general)

PR #402, opened 2023-01-11
Remove one duplicated check for glGetString(GL_VERSION)

z-s-e:
The same check is done in glad_gl_find_core_... and will return a 0 version in case of failure.


Dav1dde:
Thanks!

PR #403, opened 2023-01-13
c generator: Allow setting the function pointer field name prefix.

rpavlik:
To avoid symbol conflicts.

I don't super love my implementation here because it encodes the whole "prefix + _ + name" thing in more than one place, but it doesn't make it less DRY than it was before.

### DIFF
--- a/glad/generator/c/__init__.py
+++ b/glad/generator/c/__init__.py
@@ -111,7 +111,7 @@ def get_debug_impl(command, command_code_name=None):
 def ctx(jinja_context, name, context='context', raw=False, name_only=False, member=False):
     options = jinja_context['options']
 
-    prefix = 'glad_'
+    prefix = options['prefix'] + '_'
     if options['mx']:
         prefix = context + '->'
         if name.startswith('GLAD_'):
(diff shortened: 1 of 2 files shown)


zaruhev:
Might be necro, but I actually like this idea. However, is a default value specified for the `prefix` parameter? If not, I think it should have a default `glad` value.

Dav1dde:
Sorry I should have at least put a comment here, this is still on my "to consider" list, it is very simple so I might still end up doing it, but my stance was always it's easier/better to just ran `sed` on the generated files, this will catch all occurences.

Making the prefix configurable within the code generation just makes everything a lot more ugly, if it is done for all symbols (including API, defines etc.), but maybe just the exported symbols is enough.

zaruhev:
Hi @Dav1dde,

I think a lot of people that use glad2 write on Windows. I think this feature would be very helpful to people using systems without Unix commands. 

I'd also like to re-emphasize that I think the `prefix` parameter should be optional, such that not specifying it at all defaults to `glad` anyway. I think this would be really convenient to have in the build system especially when using CMake.

Thank you for reading

Dav1dde:
> I think this would be really convenient to have in the build system especially when using CMake.

CMake is one of the main painpoints and if there wasnt so many people using it, I would have already deleted it. You're better off including the generated files in your repo.

zaruhev:
Hi @Dav1dde,

I understand what you're getting at. Also, I sympathize with your dislike of CMake. I personally include GLAD headeronly anyway, because I don't like using it through CMake. I try to use CMake as little as possible, though I was trying to consider alternative use cases. 

Thank you for responding

PR #404, opened 2023-01-16
ci: bump actions versions

scribam:


.github/workflows/glad2.yaml
```diff
@@ -31,8 +30,8 @@ jobs:
       - name: Run Tests
         run: PRINT_MESSAGE=1 xvfb-run --auto-servernum ./utility/test.sh
       - name: Publish Test Results
-        uses: EnricoMi/publish-unit-test-result-action@v1
+        uses: EnricoMi/publish-unit-test-result-action@v2
         if: always()
         with:
-          files: test-report.xml
+          junit_files: test-report.xml
```
Dav1dde, line 36: Is this correct, the docs https://github.com/EnricoMi/publish-unit-test-result-action/tree/v2.4.1 still use `files`

scribam: Indeed, they reintroduced `files` last week https://github.com/EnricoMi/publish-unit-test-result-action/pull/395, I will update this PR accordingly

scribam: PR updated

Dav1dde: Thank you, that was amazingly fast!

PR #408, opened 2023-02-07
Ninja: Fixed "duplicate build rules" error

shamanDevel:
When building cmake-projects using ninja as generator and if multiple APIs are specified, khrplatform.h is added multiple times to the list of files to compile. This triggers an error in ninja: "duplicate build rules".
Simply removing duplicate files from the GGC_FILES list fixes this issue.


Dav1dde:
Thanks!

PR #413, opened 2023-02-20
c: correct name of unload-function

kusma:
This is named gladUnloadWGL in the header file, and the corresponding load-function also has a WGL suffix. So it seems safe to assume that the GLX suffix is a typo.


Dav1dde:
That's awkward ... thanks!

PR #418, opened 2023-03-05
Honor BUILD_SHARED_LIBS behavior

rafalmaziejuk:
When running cmake with -DBUILD_SHARED_LIBS=ON glad is not built
as shared library. This change fixes that.

Signed-off-by: rafalmaziejuk <rafalmaziejuk@gmail.com>


Dav1dde:
Thanks, I moved it to the top so the `if GG_SHARED` above is also hit

PR #429, opened 2023-06-13
Use FindPython instead of FindPythonInterp

gaboroszkar:
FindPythonInterp is deprecated. The new module finds new way to locate the python environment. In some cases the deprecated FindPythonInterp found the incorrect python interpreter.


bekogeko:
This still give an error on make using add_subdirectory.

Dav1dde:
Thanks, sorry for only getting to it so late!

PR #438, opened 2023-08-17
Improve GLES loader

halx99:
- Fix duplicated gl load handle when with both gles,gl loader
- Allow use system EGL for GLES loader


Dav1dde:
Sorry for the late reply, I implemented this in 41963067d847138c1d6e10fbad41c266a4708e47 and 08eb15c45fa2692627ce0664401a9d2595c67192. 

You can now switch to system EGL with `GLAD_GLES2_USE_SYSTEM_EGL` instead of opting out.

PR #447, opened 2023-12-20
Add argument names to function declarations

vis4rd:
This PR changes `#define`s of function declarations (effectively aliases) to function-like macros, so that at least parameter names are visible in editors like VSCode with `clangd` (which can't expand macros further than 1-level depth, apparently).

Before:
```cpp
GLAD_API_CALL PFNGLGETERRORPROC glad_glGetError;
#define glGetError glad_glGetError
GLAD_API_CALL PFNGLUNIFORM4UIPROC glad_glUniform4ui;
#define glUniform4ui glad_glUniform4ui
GLAD_API_CALL PFNGLUSEPROGRAMPROC glad_glUseProgram;
#define glUseProgram glad_glUseProgram
```

410855250_726743109039864_8198493770215275128_n

After:
```cpp
GLAD_API_CALL PFNGLGETERRORPROC glad_glGetError;
#define glGetError() glad_glGetError()
GLAD_API_CALL PFNGLUNIFORM4UIPROC glad_glUniform4ui;
#define glUniform4ui(location, v0, v1, v2, v3) glad_glUniform4ui(location, v0, v1, v2, v3)
GLAD_API_CALL PFNGLUSEPROGRAMPROC glad_glUseProgram;
#define glUseProgram(program) glad_glUseProgram(program)
```

403401978_1568828770520487_5287887910017992645_n

There is a slight possibility, that it could also prevent false-positive compilations when passed an invalid number of arguments to functions in heavily templated code. Hopefully function macros might help mitigate that.

### DIFF
--- a/glad/generator/c/templates/template_utils.h
+++ b/glad/generator/c/templates/template_utils.h
@@ -116,9 +116,9 @@ typedef {{ command.proto.ret|type_to_c }} (GLAD_API_PTR *{{ command.name|pfn }})
 GLAD_API_CALL {{ command.name|pfn }} glad_{{ command.name }};
 {% if debug %}
 GLAD_API_CALL {{ command.name|pfn }} glad_debug_{{ command.name }};
-#define {{ command.name }} glad_debug_{{ command.name }}
+#define {{ command.name }}({{ command.params|param_names }}) glad_debug_{{ command.name }}({{ command.params|param_names }})
 {% else %}
-#define {{ command.name }} glad_{{ command.name }}
+#define {{ command.name }}({{ command.params|param_names }}) glad_{{ command.name }}({{ command.params|param_names }})
 {% endif %}
 {% endcall %}
 {% endfor %}


Dav1dde:
Thanks for taking the time to make this change, unfortunately this has come up multiple times before and so far my stance on this was (and still is), IDEs just get need to get this sorted out. The first time this came up was in 2014 and IDEs like QtCreator actually resolve this correctly since 2014 or before! I am also pretty confident CLion resolves the symbols correctly.

> so that at least parameter names are visible in editors like VSCode with clangd (which can't expand macros further than 1-level depth, apparently)

From what I remember, the problem IDEs had was actually with the function pointers not even the macro. Also this is just 1 layer of macros.

The approach of tweaking the macro itself just has drawbacks. It will break a bunch of existing code among other things like making it impossible to take the address of the function pointer.

I think this is better suited as a fix in `clangd` or a switch to maybe another completion server which does not have this limitation.

PR #453, opened 2024-01-30
Provide CMake package

### LINKED ISSUE #452: Provide cmake-package
Hello,

Maintaining [conda-forge/glad-feedstock](https://github.com/conda-forge/glad-feedstock) is my goal (https://github.com/conda-forge/glad-feedstock/issues/7).

But I have an idea how to improve CMake support in upstream.
What do you think about translating [cmake/CMakeLists.txt](https://github.com/Dav1dde/glad/blob/glad2/cmake/CMakeLists.txt) into GladConfig.cmake?

To ensure backward compatibility, we can set CMAKE_PREFIX_PATH to a new location of GladConfig.cmake and enable the use of python module from sources inside cmake/CMakeLists.txt.

If it's okay with you, I can provide PR.

cNoNim:
Closes #452


cNoNim:
Create draft PR.
Will test.

cNoNim:
I tested on Windows and Linux.

tested by installing the module via
```shell
pip install.
```
+ manually copying GladConfig.cmake to $PREFIX/lib/cmake/Glad

```cmake
find_package(Glad REQUIRED)

glad_add_library(glad_gl_core_33 REPRODUCIBLE API gl:core=3.3)
```

Also tested by
manual copy sources to ${PROJECT_SOURCE_DIR}/external/glad/ of sample project

```
set(GLAD_SOURCES_DIR "${PROJECT_SOURCE_DIR}/external/glad/")
add_subdirectory("${GLAD_SOURCES_DIR}/cmake" glad_cmake)

glad_add_library(glad_gl_core_33 REPRODUCIBLE API gl:core=3.3)
```

All works

cNoNim:
> When using the following CMake snippet in my local project, CMake will fail saying it can't find `glad_add_library`. Though, interestingly enough, this only happens for the CMake configurations which use MSVC as the compiler.
> 
> ```cmake
> add_subdirectory("${FASTGLTF_DEPS_DIR}/glad/cmake" glad_cmake)
> 
> # Let the glad script generate the glad headers
> glad_add_library(fg_glad_gl46 REPRODUCIBLE EXCLUDE_FROM_ALL LOADER API gl:core=4.6)
> ```
> 
> However, if I change the `cmake/CMakeLists.txt` file like this, it works again as it did before.
> 
> ```diff
> -list(APPEND CMAKE_PREFIX_PATH ${GLAD_SOURCES_DIR})
> -find_package(Glad REQUIRED)
> +include(./GladConfig.cmake)
> ```
> 
> Should I make this into an issue or provide the PR to fix this? Note that I am on Windows using CLion and CMake 3.27.8.

Consider opening an issue.
The logic described in the wiki suggests the need to set GLAD_SOURCES_DIR, but it may not be necessary.
If it is unnecessary, it can be fixed as you suggest. Keep in mind that GLAD_SOURCES_DIR defines the working directory.
> "Directory containing glad sources (python modules), used as working directory. Must be absolute."

cNoNim:
```cmake
set(GLAD_SOURCES_DIR "${FASTGLTF_DEPS_DIR}/glad")
add_subdirectory("${GLAD_SOURCES_DIR}/cmake" glad_cmake)
```
I assume this is how it will work.

cNoNim:
I now understand why everything worked without specifying the path.
In this case, replacing 'find_package' with 'include' seems correct.
I apologize for the regression.

spnda:
> ```cmake
> set(GLAD_SOURCES_DIR "${FASTGLTF_DEPS_DIR}/glad")
> add_subdirectory("${GLAD_SOURCES_DIR}/cmake" glad_cmake)
> ```
> 
> I assume this is how it will work.

That would not have made any difference because glad's `CMakeLists.txt` defines it too. So setting it or not setting it should make absolutely no difference as far as I can tell.

> I now understand why everything worked without specifying the path.
In this case, replacing 'find_package' with 'include' seems correct.
I apologize for the regression.

Shall I open a pull request?

cNoNim:
I thought that `find_package` and `CMAKE_PREFIX_PATH` were the same as `include`. 
However, exist issue https://gitlab.kitware.com/cmake/cmake/-/issues/16644.
Therefore, using `include` is more consistent with the previous behavior.

I don't mind PR. I guess @Dav1dde feels the same way.

And thanks.

Issue #463, opened 2024-04-02
GL/GLFW Example Displays GLAD Version "0.1"


zaruhev:
The GL/GLFW example located at `example/c/gl_glfw.c` (shown below) loads GL with `glfwGetProcAddress`:
```c
int version = gladLoadGL(glfwGetProcAddress);
```

However, when loading with GLFW (and glad2) for OpenGL 4.6, `gladLoadGL(GLADloadfunc)` returns `1`. Thus, when retrieving the major and minor versions of OpenGL, `"0.1"` is generated.

I'm not sure if this is intentional, because I would expect this to return either the current GLAD version or the current GL version, however only `1` is ever returned from `gladLoadGL`.

I don't think this issue is cause for any immediately obvious problems with the loader, but I worry about portability in the future. Is this the incorrect way to retrieve the OpenGL version loaded by glad? If so, what should be done instead?

Thank you for reading.

Contents of `example/c/gl_glfw.c`

```c
#include <glad/gl.h>
// GLFW (include after glad)
#include <GLFW/glfw3.h>

#include <stdlib.h>
#include <stdio.h>


const GLuint WIDTH = 800, HEIGHT = 600;

void key_callback(GLFWwindow* window, int key, int scancode, int action, int mode) {
    if (key == GLFW_KEY_ESCAPE && action == GLFW_PRESS)
        glfwSetWindowShouldClose(window, GL_TRUE);
}

int main(void) {
    glfwInit();

    glfwWindowHint(GLFW_CONTEXT_VERSION_MAJOR, 4);
    glfwWindowHint(GLFW_CONTEXT_VERSION_MINOR, 6);
    glfwWindowHint(GLFW_OPENGL_PROFILE, GLFW_OPENGL_CORE_PROFILE);

    GLFWwindow* window = glfwCreateWindow(WIDTH, HEIGHT, "[glad] GL with GLFW", NULL, NULL);
    glfwMakeContextCurrent(window);

    glfwSetKeyCallback(window, key_callback);

    int version = gladLoadGL(glfwGetProcAddress);
    printf("GL %d.%d\n", GLAD_VERSION_MAJOR(version), GLAD_VERSION_MINOR(version)); // <-- prints "GL 0.1"

    while (!glfwWindowShouldClose(window)) {
        glfwPollEvents();

        glClearColor(0.7f, 0.9f, 0.1f, 1.0f);
        glClear(GL_COLOR_BUFFER_BIT);

        glfwSwapBuffers(window);
[...]

PR #464, opened 2024-04-03
Defined GLAD_GL_IMPLEMENTATION to GLFW example

zaruhev:
Patch for issue #463 which can cause portability issues in the future by defining `GLAD_GL_IMPLEMENTATION` before glad import in GLFW example.

This patch does not stress nor attempt to fix any issues with the library itself, but edits the example in question so consumers don’t make the mistake of not defining the required macro.

> An alternative solution might be to include GLFW before glad, but I haven’t had the chance to confirm. Don’t take my word for it.

Also added .vscode to .gitignore. If this is
an issue please let me know so I can remove
it before the pull.

Thank you @Dav1dde for helping me troubleshoot the version checking issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ dist/
 /rust/
 target/
 Cargo.lock
+.vscode/
--- a/example/c/gl_glfw.c
+++ b/example/c/gl_glfw.c
@@ -1,10 +1,11 @@
+#include <stdlib.h>
+#include <stdio.h>
+
 #define GLAD_GL_IMPLEMENTATION // Necessary for headeronly version.
 #include <glad/gl.h>
 
 #include <GLFW/glfw3.h>
 
-#include <stdlib.h>
-#include <stdio.h>
 
 const GLuint WIDTH = 800, HEIGHT = 600;
 
--- a/example/c/gl_glfw.c
+++ b/example/c/gl_glfw.c
@@ -1,6 +1,7 @@
 #include <stdlib.h>
 #include <stdio.h>
 
+#define GLAD_GL_IMPLEMENTATION
 #include <glad/gl.h>
 
 #include <GLFW/glfw3.h>
--- a/example/c/gl_glfw.c
+++ b/example/c/gl_glfw.c
@@ -1,7 +1,6 @@
-#define GLAD_GL_IMPLEMENTATION
+#define GLAD_GL_IMPLEMENTATION // Necessary for headeronly version.
 #include <glad/gl.h>
 
-#define GLFW_INCLUDE_NONE
 #include <GLFW/glfw3.h>
 
 #include <stdlib.h>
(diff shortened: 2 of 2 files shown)


zaruhev:
Sorry, finger slipped

Dav1dde:
Thanks for the patch, looks good, also adding vscode is fine! I am a bit surprised though, glad should return `0` in this case not `1`. I'll have to dig into why it returns `1`.

zaruhev:
Hi @Dav1dde,

To my knowledge and experience using GLAD for a couple years, I assumed the `gladLoadGL` return code was supposed to be a success indicator. At least, I've always used it that way, as have examples that use your code. It would be funny if the root of this was a misconception, but it seems fine. Unless I am misunderstanding the bug?

Also, I don't have merge permissions, so if the patch is sufficient would you mind merging when you have the chance?

Thank you,
Sarah

zaruhev:
Oh, I just understood what you meant. If I get the chance I'll look into why glad is being set when the version is not set.

It could be that the loader runs anyway whilst the version checker doesn't? I'll take a gander when I have the chance. Thanks for the insight.

zaruhev:
Hi,

Quick follow up: `GLFW_INCLUDE_NONE` should be included before importing GLFW3 in order to avoid redeclarations of the OpenGL function macros.

Thank you for reading.

Dav1dde:
> Quick follow up: I updated the fork, because GLFW_INCLUDE_NONE should be included before importing GLFW3 in order to avoid redeclarations of the OpenGL function macros.

This is not necessary as long as you include glad **before** glfw.

> To my knowledge and experience using GLAD for a couple years, I assumed the gladLoadGL return code was supposed to be a success indicator.

It is, the previous version used `0` and `1`, glad2 uses `0` and `some value bigger than 0 but not 1`, which means the boolean check still works but you can also extract the loaded OpenGL version from it. Hence why glad shouldn't ever return `1`.

> Also, I don't have merge permissions, so if the patch is sufficient would you mind merging when you have the chance?

I will, but it might be open for a few days, I want to take the opportunity to figure out why glad returns `1` in the first place.

zaruhev:
Hi @Dav1dde,

I see. Thank you for the follow up. I was trying to reproduce the return 1 issue but I seemed to have failed to documented my reproduction steps. I used the headeronly library when troubleshooting and I'm using it now, but when I undefine `GLAD_GL_IMPLEMENTATION` it (rightfully) fails to link. I haven't figured out how to reproduce the issue.

Also, I checked `glfw3.h` again, and confirmed your mentioning that `GLFW_INCLUDE_NONE` is not necessary when `__gl_h_` is defined. Sorry for that oversight, I'll revise this in my next commit.

Thank you.

zaruhev:
Good morning @Dav1dde,

I'm thinking the same thing must have happened on my end, as well. I could not reproduce the issue, but in trying to do so I actually learned a lot about the library :)

Thanks for keeping up with this branch/issue

PR #470, opened 2024-05-17
Fix `UnicodeDecodeError` in non-UTF8 locale

### LINKED ISSUE #469: Incrroect encoding when python read file
https://github.com/Dav1dde/glad/blob/1ecd45775d96f35170458e6b148eb0708967e402/glad/files/__init__.py#L22-L31

If `resource_exists` failed, script will use `open` to read file, but local host encoding. So I got the error:

```
UnicodeDecodeError: 'gbk' codec can't decode byte 0xbf in position 2: illegal multibyte sequence
```

Here are some of my suggestions.

- It's not recommended to use `pkg_resources`.

https://setuptools.pypa.io/en/latest/pkg_resources.html
> Use of pkg_resources is deprecated in favor of importlib.resources, importlib.metadata and their backports (importlib_resources, importlib_metadata). Some useful APIs are also provided by packaging (e.g. requirements and version parsing). Users should refrain from new usage of pkg_resources and should work to port to importlib-based solutions.

- Use `PYTHONWARNDEFAULTENCODING=1` to fix potential encoding errors.

GalaxySnail:
fix #469

### DIFF
--- a/glad/files/gl.xml
+++ b/glad/files/gl.xml
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="UTF-8"?>
+<?xml version="1.0" encoding="UTF-8"?>
 <registry>
     <comment>
 Copyright 2013-2020 The Khronos Group Inc.
--- a/glad/__main__.py
+++ b/glad/__main__.py
@@ -40,7 +40,7 @@ def main():
         if reproducible:
             logger.info('reproducible build, using packaged specification: \'%s.xml\'', value)
             try:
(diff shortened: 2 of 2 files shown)


glad/files/gl.xml
```diff
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="UTF-8"?>
```
Dav1dde, line 1: The change should also work with the BOM, since the files are just dumped directly from the git without any (manual) processing.

I tested it locally and it seems to work fine with the BOM, can you revert this change?

Issue #247, opened 2020-02-28
Include doc comments


PatrikValkovic:
The Khronos reference page contains description of parameters and description of the behavior for every function. Wouldn't it be possible to add doc comments for the functions, so IDE can show it? I understand the function description is a bit cumbersome, as one description is usually related to multiple functions with similar functionality, however parameters description would help a lot.
I am completely new to OpenGL and it would be helpful to me if IDE hinted me what each parameter means and I didn't need to search for every function.

Dav1dde:
I was messing around with the docs a while ago: https://github.com/Dav1dde/glad-guru using https://github.com/KhronosGroup/OpenGL-Refpages

Unfortunately the docs are not easily processable and properly supporting such a feature just takes too much time and effort I don't have right now.

If there is a relatively sane documentation database, I am happy to include it in glad and (optionally) generate docstrings from it.

PatrikValkovic:
Is there any way how I could help?
I think just the main description (in the name section) and parameters description can be enough for the beginning.

Dav1dde:
The problem is the [spec](https://github.com/KhronosGroup/OpenGL-Registry/blob/master/xml/gl.xml) holds no documentation, so we need a different source.

As far as I am aware there is no good, easily machine readable documentation. The above linked Refpages are okay but need quite a bit of post processing.

So for now there isn't much you can do, you can work on a (preferably Python) library that parses the refpages (OpenGL/Vulkan) into a machine usable format.

dimateos:
Sorry for commenting on an old thread, just wondering if you still think the same.

Maybe you can use https://github.com/BSVino/docs.gl direclty.

Thanks for glad :)

PR #471, opened 2024-05-18
Code documentation support

paoloose:
This PR adds support for optional code documentation as requested in #247.

Basically, `Specification` classes can now include documentation from a language-agnostic `SpecificationDocs` class, which knows how to retrieve and parse the specs documentation from a specific source.

As a proof of concept I've implemented the `DocsGL(SpecificationDocs)` class for the `GL` specification, using <https://docs.gl/> as source. docs.gl was definitely not created to be parser-friendly, but I managed to properly parse all the commands docs (without including tables and code blocks for now).

Additional sources will be needed for others specifications. Currently I'm only planning to contribute with the GL spec docs. Others can be added later with no problems.

Looking forward to feedback and willing to implement any requested modification. Thanks!

Test with:

```bash
# Use the --with-docs flag
python -m glad --extensions="" --api="gl:compatibility=" --out-path ./temp c --with-docs
```

Thinks to know:

- Currently, only commands are documented.
- Documentation code is optional and included only when the `--with-docs` flag is present. If the given specification doesn't support docs, only a warning is shown.
- Documentation was only added to C templates, but can be easily added to Rust templates also.


paoloose:
image

This is how hovering a command looks in vscode

glad/documentation.py
```diff
@@ -0,0 +1,171 @@
...
+import glad.util
+import subprocess
+import re
+
+
+class DocsGL(SpecificationDocs):
+    URL = 'https://github.com/BSVino/docs.gl.git'
+    CACHED = True # Only clones the DocsGL repo once
+    API = 'gl'
+    docs = dict()
```
Dav1dde, line 13: I think you wanted this on the instance not class?

glad/documentation.py
```diff
@@ -0,0 +1,171 @@
...
+
+    @property
+    @memoize(method=True)
+    def out_dir(self):
+        from pathlib import Path
+        from tempfile import gettempdir
+        if self.CACHED:
+            return Path('.cached') / 'docs.gl'
+        return Path(gettempdir()) / 'docs.gl'
```
Dav1dde, line 22: This should either be a field on the instance or just a temporary variable.

glad/documentation.py
```diff
@@ -0,0 +1,171 @@
...
+    URL = 'https://github.com/BSVino/docs.gl.git'
+    CACHED = True # Only clones the DocsGL repo once
+    API = 'gl'
+    docs = dict()
+
+    @property
+    @memoize(method=True)
+    def out_dir(self):
+        from pathlib import Path
+        from tempfile import gettempdir
```
Dav1dde, line 19: Please all imports at the top.

glad/documentation.py
```diff
@@ -0,0 +1,171 @@
...
+        if self.CACHED:
+            return Path('.cached') / 'docs.gl'
+        return Path(gettempdir()) / 'docs.gl'
+
+    def load(self):
+        if self.out_dir.exists() and not self.CACHED:
+            rmtree(str(self.out_dir))
+        if not self.out_dir.exists():
+            subprocess.run(['git', 'clone', '--depth=1', self.URL, str(self.out_dir)])
```
Dav1dde, line 28: Ideally we download a zip

glad/documentation.py
```diff
@@ -0,0 +1,171 @@
...
+    def out_dir(self):
+        from pathlib import Path
+        from tempfile import gettempdir
+        if self.CACHED:
+            return Path('.cached') / 'docs.gl'
+        return Path(gettempdir()) / 'docs.gl'
+
+    def load(self):
+        if self.out_dir.exists() and not self.CACHED:
+            rmtree(str(self.out_dir))
```
Dav1dde, line 26: I feel like the caching behaviour should be passed into the class/load function. Deleting files should be done only on the cli. E.g. this won't behave very nice on the webservice.

glad/documentation.py
```diff
@@ -0,0 +1,171 @@
...
+        if e.tag == 'mfenced':
+            return f'{e.attrib["open"]}{e.text}'
+        if e.tag == 'dt':
+            return f'\n{CommandDocs.BREAK}- '
+        if e.tag == 'li':
+            return f'\n{CommandDocs.BREAK}-{e.text}'
+        return e.text
+
+    @staticmethod
+    def xml_text(e):
```
Dav1dde, line 165: Could just be a normal free standing function

glad/parse.py
```diff
@@ -1,3 +1,4 @@
+from tkinter import N
```
Dav1dde, line 1: Leftover? Definitely don't want to depend on tkinter

glad/parse.py
```diff
@@ -654,7 +669,7 @@ def split_types(iterable, types):
 
         return result
 
-    def select(self, api, version, profile, extension_names, sink=LoggingSink(__name__)):
+    def select(self, api, version, profile, extension_names, config, sink=LoggingSink(__name__)):
```
Dav1dde, line 672: This shouldn't directly depend on the config, better if this was just a kwarg

glad/parse.py
```diff
@@ -1453,3 +1478,52 @@ def from_element(cls, element, api=None):
...
+
+
+class CommandDocs(object):
+    """
+    Inline code documentation for a command/function.
+    """
+
+    # Template rendering will interpret this as a custom paragraph break.
+    # If no special break logic is needed, just use '\n'.
+    BREAK = '__BREAK__'
```
Dav1dde, line 1509: I think it's better if instead we just had a list of paragraphs instead of joining and then splitting it after again.

glad/parse.py
```diff
@@ -1453,3 +1478,52 @@ def from_element(cls, element, api=None):
     def __str__(self):
         return '{self.name}@{self.version!r}'.format(self=self)
     __repr__ = __str__
+
+
+class SpecificationDocs(object):
```
Dav1dde, line 1483: I think this fits better in the documentation module

glad/parse.py
```diff
@@ -1453,3 +1478,52 @@ def from_element(cls, element, api=None):
...
+    def __init__(self, brief, params, description, notes, errors, see_also):
+        self.brief = brief
+        self.params = params
+        self.description = description
+        self.notes = notes
+        self.errors = errors
+        self.see_also = see_also
+
+    def __str__(self):
+        return 'DocComment(brief={!r}, ' \
```
Dav1dde, line 1523: Class is called `CommandDocs`.

```suggestion
        return 'CommandDocs(brief={!r}, ' \
```

glad/parse.py
```diff
@@ -732,6 +747,15 @@ def select(self, api, version, profile, extension_names, sink=LoggingSink(__name
         extensions = [self.extensions[api][name] for name in extension_names
                       if name in self.extensions[api]]
 
+        # Load documentation for the specific configuration
+        if config['WITH_DOCS']:
```
Dav1dde, line 751: Seems like the specification doesn't need the docs at all. Instead of tying the docs to the specification, the templating could be the part where the specification gets married with the docs.

paoloose:
> I wouldn't want to depend on docs.gl but instead the Khronos repo [OpenGL-Refpages](https://github.com/KhronosGroup/OpenGL-Refpages) (except if there is a good reason to depend on docs.gl).

Seems good to me. I'll be replacing DocsGL with that source. I understand that docs.gl is more a community driven documentation than a strict reference. I was also consider shipping both, but from a maintainer point of view docs.gl is not convenient.

Working on resolving the other requested changes :)

glad/parse.py
paoloose: Yeah, probably auto imported by my ide haha

paoloose: Yeah, probably auto-imported by my ide lol

glad/parse.py
paoloose: Paragraphs are list of strings joined by a like break. However I needed another level of "breaking" logic, for example, for list items, term definitions, etc. I'll try to remove this logic.

glad/parse.py
paoloose: I based the `documentation,py` module on how the `specification.py` is structured: just classes that inherit from the base `Specification` class defined in `parse.py`. So `documentation.py` holds specialized classes that inherit from the base `SpecificationDocs` class, also defined in `parse.py`.

glad/parse.py
Dav1dde: I am not quite sure about how I'd like it to look at the end, don't worry too much about it, it's easy for me to just move everything a bit around.

But basically in mind right now the docs are a separate source of data independent of the spec and you just pass it as some kind of "docs db" into the template and the template tries to lookup the docs from the db when rendering.

What guarantuess you get the correct symbols? Either it's in the template, a similar mechanism to `spec.select()` or just nothing, if you pass the wrong docs then there are just no docs for the requested symbols.

But again don't worry too much about that, these are the things that are easy to change later, the hard part is the actual logic!

Dav1dde:
> I understand that docs.gl is more a community driven documentation than a strict reference. 

I assumed it is based on the same OpenGL-Refpages repo just already processed for better display on docs.gl but without additional curation.

glad/parse.py
Dav1dde: I am not too worried about, I was just thinking instead of keeping the breaks in the text, we can just keep a list of text instead.

glad/documentation.py
paoloose: Done

glad/documentation.py
paoloose: Done

glad/documentation.py
paoloose: Removed and move this logic to `__main__.py`, like specification does

glad/documentation.py
paoloose: Removed, now I use a `DocumentationSet` similar to `FeatureSet`

glad/documentation.py
paoloose: I remove this caching logic for now, and move it to `__main__.py` similar to how specifications are loaded (from file if exist, otherwise from remote)

glad/documentation.py
paoloose: Do you mean an instance method? If so, I can't since I need this in `docs_from_html_file`, which is a `@classmethod` function

glad/parse.py
paoloose: I moved this logic and now docs are loaded separately, only for the specifications listed by the user. Then, those loaded docs sources are "selected" from a feature_set, and passed directly to the template generator function, without depending on the spec. I think I ended up understanding what you meant, but if not, feel free to change it later. I'll be working on using the Khronos Refpages as source :-)

paoloose:
> I assumed it is based on the same OpenGL-Refpages repo just already processed for better display on docs.gl but without additional curation.

See <https://docs.gl/about.html>. It adds additional information to the official specs (even typo fixes) and allows contributions, but yeah, what was said about maintenance still holds. Still working on parsing Khronos reference.

glad/documentation.py
```diff
@@ -0,0 +1,190 @@
+import re
+import glad.util
+from lxml import etree
```
Dav1dde, line 3: Does this work without lxml as well? Ideally lxml is optional and just a performance improvement but not necessary.

glad/documentation.py
```diff
@@ -0,0 +1,190 @@
...
+from glad.util import prefix, suffix, raw_text
+
+class OpenGLRefpages(SpecificationDocs):
+    DOCS_NAME = 'opengl_refpages'
+
+    URL = 'https://github.com/KhronosGroup/OpenGL-Refpages/archive/refs/heads/main.zip'
+    SPEC = 'gl'
+
+    def select(self, feature_set):
+        current_major = list(feature_set.info)[0].version.major
```
Dav1dde, line 14: When merging multiple specs this can contain multiple infos of all the merged versions.

Seems like this should work here:
```python
max(info.version.major for info in feature_set.info)
```

glad/parse.py
```diff
@@ -1453,3 +1445,90 @@ def from_element(cls, element, api=None):
...
+        zip_out_dir = docs_dir.parent
+
+        with closing(opener.urlopen(url)) as f:
+            raw = f.read()
+            f = BytesIO(raw)
+
+            with zipfile.ZipFile(f) as zf:
+                zf.extractall(zip_out_dir)
+                zip_dir = zip_out_dir / zf.namelist()[0]
+                os.rename(zip_dir, docs_dir)
```
Dav1dde, line 1482: I'd prefer if the docs worked with a zipfile without extracting it. Should be able to traverse the zip instead of the filesystem.

glad/parse.py
```diff
@@ -1,35 +1,27 @@
...
-except ImportError:
-    try:
-        import xml.etree.cElementTree as etree
-    except ImportError:
-        import xml.etree.ElementTree as etree
-
-    def xml_fromstring(argument):
-        return etree.fromstring(argument)
-    def xml_parse(path):
-        return etree.parse(path).getroot()
```
Dav1dde, line 20: This was here for compatibility when there is no lxml installed.

glad/__main__.py
```diff
@@ -71,6 +71,11 @@ class GlobalConfig(Config):
         description='Makes the build reproducible by not fetching the latest '
                     'specification from Khronos.'
     )
+    WITH_DOCS = ConfigOption(
```
Dav1dde, line 74: I think I prefer `docs` here.

```suggestion
    DOCS = ConfigOption(
```

Dav1dde (commented):
Thanks for putting in all the work and sorry for being so unresponsive slow, I just don't have the mental bandwidth atm.

glad/parse.py
paoloose: Done. I reverted this as it was. The reason I initially dropped xml support was because it's parser lacked [features](https://lxml.de/api/lxml.etree.XMLParser-class.html) such as recovery (now managed by injecting namespaces where needed), entities resolution (which I now handle manually for better control), and the ability to retrieve a node's parent (which I also now do manually).

Now both the lxml and xml modules are supported and produce identical results.

glad/parse.py
paoloose: Done :), didn't know it was possible

glad/documentation.py
paoloose: Done. Now it works with both lxml and xml

paoloose:
> Thanks for putting in all the work and sorry for being so unresponsive slow, I just don't have the mental bandwidth atm.

No worries, thank you for taking the time to review it. I'm dogfooding this docs while I'm working in some OpenGL projects, so I'll be pushing fixes as I encounter more issues. For now I've implemented the requested changes. Wish the code was more elegant, but Khronos really worked hard to make their documentation format a parsing nightmare haha.

MrLuigiBean:
Hello, apologies for reviving this old thread, but are the changes from this branch going to be added to the default branch? I'd love to have documentation generated and visible when hovering over functions to make development a little easier.

eloj:
Just an update on the OpenGL-RefPages situation.

Khronos (perhaps understandably) hasn't been paying too much attention to the repo, but I pinged Jon Leech of Khronos last week, and he got the GL 4.6 pages and many, many other pending pull requests [merged](https://github.com/KhronosGroup/OpenGL-Refpages/commits/main/).

There are still a couple PRs ([glGet args](https://github.com/KhronosGroup/OpenGL-Refpages/pull/146), [min tex sizes](https://github.com/KhronosGroup/OpenGL-Refpages/pull/144)) with fixes pending, but now the original source is much more up-to-date.

(BTW: I added GLAD to the OpenGL-RefPages tools list on the [index page](https://registry.khronos.org/OpenGL-Refpages/))

eloj:
Anyway, I have tested this branch against the updated specification, and it seems to work after the following changes:

* A rebase is needed. The only conflict is in `.gitignore`, so easily resolved.
* One of the XML files (`gl4/glPolygonOffsetClamp.xml`) now contains an `&lcub;` entity, so something like the following patch is required.

```diff
diff --git a/glad/util.py b/glad/util.py
index bc5ca54..ff07843 100644
--- a/glad/util.py
+++ b/glad/util.py
@@ -267,6 +267,8 @@ math_symbols_map = {
     '&RightCeiling;': '⌉',
     '&DoubleVerticalBar;': '∥',
     '&VerticalBar;': '|',
+    '&lcub;': '{',
+    '&rcub;': '}',
 }
```

Probably the handling of entities could be made more robust.

eloj:
The main deficiency I see with this branch at this point, is that the `--docs` option is not reflected in generated template output (e.g `include/glad/gl.h`).

I can see that 'docs' is not present in the `gen_info` in the `Commandline.build()` method, but honestly after 20 minutes of trying to figure how this is supposed to flow, I gave up. For now.

paoloose:
Thanks for putting in the work! I've implemented the two changes you mentioned, but don't currently have the time to check the `--docs` bug

Happy to see this branch getting some attention again :)

Issue #466, opened 2024-04-13
GCC -fanalyzer report error when compiling glad


Raffaello:
file: glad.c
line: 210
error: `use of uninitialized value '*exts_i.20_2 + _4' [CWE-457] [-Werror=analyzer-use-of-uninitialized-value]`

glad generated 4.3 core.

lephilousophe:
I can't reproduce this error with a GCC 13.2.1.
Can you provide more detailed instructions?

Dav1dde:
Thanks for checking, I can't either. Please re-open if you have something we can reproduce.

Raffaello:
apologies i forgot to mention: This is due to GCC 14,
Fedora 40 has GCC 14.

lephilousophe:
I still cannot reproduce using Fedora 40 `g++ (GCC) 14.1.1 20240522 (Red Hat 14.1.1-4)`

With gl.h from http://glad.sh/#api=gl%3Acore%3D4.3&generator=c&options=ALIAS%2CHEADER_ONLY%2CLOADER and the following glad.c:
```
#define GLAD_GL_IMPLEMENTATION

#include "gl.h"

int main() {
    gladLoaderLoadGL();
    return 0;
}
```

And compiling with `g++ -o /tmp/glad /tmp/glad.c -fanalyzer`.

Raffaello:
apologies again, the error is a warning by default.

so enabling these switches `-Wall -Wextra` and enabling warning as error `-Werror`
with `-fanalyzer`

it should generate an output like this one:

```
glad.c:233:13: error: use of uninitialized value '*exts_i.20_2 + _4' [CWE-457] [-Werror=analyzer-use-of-uninitialized-value]
[build]   233 |             free((char*) exts_i[index]);
[build]       |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~
[build]   'find_extensionsGL': events 1-2
[build]     |
[build]     | 1538 | static int find_extensionsGL(void)
[build]     |      |            ^~~~~~~~~~~~~~~~~
[build]     |      |            |
[build]     |      |            (1) entry to 'find_extensionsGL'
[build]     | 1539 | {
[build]     | 1540 |     if (!get_exts())
[build]     |      |          ~~~~~~~~~~
[build]     |      |          |
[build]     |      |          (2) calling 'get_exts' from 'find_extensionsGL'
[build]     |
[build]     +--> 'get_exts': events 3-12
[build]            |
[...]

PR #473, opened 2024-06-07
Fix issue #466

lephilousophe:
Make types consistent as it troubles the GCC analyzer.

This will fix issue #466 (tested on a stripped version of generated file on Godbolt).


Dav1dde:
Thanks!

PR #483, opened 2024-08-14
Use a base function pointer instead of a regular pointer

kenan-kartal:
Passing the function pointers with regular pointers result in undefined behaviour. Before this commit, UBSan Clang Sanitizer catches this issue.

I think casting function pointers back to their original signatures from the base pointer does not result in undefined behaviour. At least UBSan does not complain after these changes.

This commit should not affect existing dependants because it introduces a single type `GLADbasefp` and it's used to change the return type of existing API only. Better test it though!

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,8 @@ struct gladGLversionStruct {
 
 extern struct gladGLversionStruct GLVersion;
 
-typedef void* (* GLADloadproc)(const char *name);
+typedef void (* GLADbasefp)(void);
+typedef GLADbasefp (* GLADloadproc)(const char *name);
 
 /*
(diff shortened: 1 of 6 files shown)


Dav1dde:
This is actually undefined behaviour for all platforms for which casting a function pointer from/to a `void *` pointer is not defined/correct, but since OpenGL itself relies on this behaviour, it essentially makes it non existent behaviour, you wouldn't use glad on a platform where OpenGL (or any graphics API) does not exist.

But agreed this is not optimal and glad could be better here and I changed this for glad2 (please confirm it is fixed there!). But for glad1 I will not change the behaviour now since this is a breaking change and will require users to add casts.

PR #498, opened 2024-11-09
Create venv and install glad-python dependencies during build

chill-nemesis:
I was playing around with the automatic generation of sources during the build process and noticed that python-glad requires jinja2, a dependency that I have not (per default) installed on my system.

With this PR I propose that, during build, a virtualenv is created and all python dependencies are installed.

I did some minor testing on my windows machine with msys and vs community 22 (amd64), all running cmake version 3.30.5

### DIFF
--- a/cmake/GladConfig.cmake
+++ b/cmake/GladConfig.cmake
@@ -241,6 +241,16 @@ function(glad_add_library TARGET)
     string(REPLACE "${GLAD_DIR}" GLAD_DIRECTORY GLAD_ARGS_UNIVERSAL "${GLAD_ARGS}")
     set(GLAD_ARGS_PATH "${GLAD_DIR}/args.txt")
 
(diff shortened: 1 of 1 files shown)


chill-nemesis:
I found a minor bug in my setup regarding the venv path. If the host is running windows, but uses e.g. msys, the venv interpreter needs to be unix style rather than windows. I'm writing a fix and update the PR.

Dav1dde:
I think it's more in the interest of the user to setup the necessary Python env which then can be picked up by glad's cmake. This has the upside, if they already need a venv, it can just be re-used.

The approach of generating a venv first seems very error prone (different environments, different Python versions, read only filesystem paths) and quite intrusive.

If you don't want to setup/install Python at all, just include the generated files into your source directory and re-generate on demand manually.

chill-nemesis:
Fair point. Then I suggest an update to the wiki that others do not run into the same issue.

PR #504, opened 2025-01-25
Close Win32 Window Properly in WGL Example

carsonalh:
The created window in this example would not close when the 'x' button was pressed.  Fixed the handling of WM_CLOSE and WM_QUIT to be DestroyWindow(hWnd) instead of PostQuitMessage(0).

Source: https://learn.microsoft.com/en-us/windows/win32/learnwin32/closing-the-window .


Dav1dde:
Thanks!

PR #510, opened 2025-04-05
Fix double space in date

1vanK:
Before: `Sat Apr  5 06:29:26 2025`
After: `Sat Apr 5 06:29:26 2025`


Dav1dde:
Thanks, but that's not a bug, just the local representation for single digit dates.

1vanK:
And why shouldn't this be fixed?

PR #522, opened 2025-08-26
Added documentation generation for C

gusg21:
Hello! I added Doxygen-style comment generation for C, I found I needed it for VS Code because it won't resolve the macro and show me the actual types. I bet there's better ways to implement this within Glad, but this worked for my needs and I thought I would at least put this out there.


gusg21:
NVM, this looks like a duplicate. Oops!

Issue #177, opened 2018-10-01
glad2: gladLoaderLoadEGL unable to load the correct EGL version.


22427:
Good Evening,
I am using glad2 in a small project, testing its awesome new features. When using it to initialize EGL I stumbled over the following issue. Compiling the minimal example shown below, only EGL 1.0 features are loaded, however the version string states, that EGL 1.4 is supported.
I think this is due to a problem in the loader function. `eglQueryString` always returns `NULL` if `eglInitialize` was not called before, causing the loader in to load only for EGL1.0. (see `static int glad_egl_find_core_egl(EGLDisplay display) `)
To call `eglInitialize` however, `gladLoaderLoadEGL` must be called first, to populate the `eglInitialize` function pointer, which leads to a bit of a chicken-and-egg  problem (?).
I am relatively new to the whole EGL topic, but if my understanding is correct I think a  feasible fix would be, to add `eglInitialize` and `eglTerminate` to the "manually" loaded function pointers in `gladLoadEGLUserPtr`. 
This would allow us to call `eglInitialize` and `eglTerminate` in `glad_egl_find_core_egl`, making the `eglQueryString` call obsolete in the process. 
I attached my manually fixed egl.h as well as the example program to this issue.

Best regards
22427

[glad2_issue.zip](https://github.com/Dav1dde/glad/files/2434686/glad2_issue.zip)

Example Code:
```C
#include <stdio.h>
#include <X11/Xlib.h>

#define GLAD_EGL_IMPLEMENTATION
#if 1
#include "glad/egl.h"
#else
#include "glad/egl_fixed.h" //<< the fixed version
#endif


int main()
{
	int ver = 0;
	EGLDisplay display = EGL_NO_DISPLAY;
	Display* xdisplay = XOpenDisplay(0);

	if(!(ver=gladLoaderLoadEGL(NULL))) {
		printf("gladLoaderLoadEGL failed!\n");
	} else {
		printf("glad:\t%d.%d\n",
		       GLAD_VERSION_MAJOR(ver),
		       GLAD_VERSION_MINOR(ver));
	}
	display = eglGetDisplay(xdisplay);

	if(display == EGL_NO_DISPLAY){
		printf("eglGetDisplay failed!\n");
	}
	if(!eglInitialize(display,nullptr,nullptr)){
		printf("eglInitialize failed!\n");
	}
	printf("EGL:\t%s\n",eglQueryString(display,EGL_VERSION));
	return  0;
}
```
Output:
```
[...]

PR #531, opened 2026-01-12
Clarify EGL loading steps with comments

ritchielrez:
Added comments to clarify EGL loading process. Moreover, mentioned issue https://github.com/Dav1dde/glad/issues/177 for better context. Normally without comments, the double loading of EGL may go unnoticed by new users of glad.


Dav1dde:
Thanks!

PR #532, opened 2026-01-31
Add SDL3 example.

eloj:
Deliberately minimized the diff against the SDL2 example, and did not change to callback-based main.


Dav1dde:
Awesome, thanks! 

> and did not change to callback-based main

If you want, I'd happy to merge a second example with a callback based approach, `example/c/gl_sdl3_callback.c` or similar.

eloj:
Okay, I'll see what I can cook up.

PR #533, opened 2026-01-31
example: adds SDL3 example (callback main)

eloj:
As per #532, here's a SDL3 example using [the callback interface](https://wiki.libsdl.org/SDL3/README-main-functions#main-callbacks-in-sdl3).

It's based on the official "[01-clear](https://examples.libsdl.org/SDL3/renderer/01-clear/)" example, but I've cut out some over the most overbearing comments.

I did leave the color-cycling in.


Dav1dde:
Awesome, thanks a lot!

PR #536, opened 2026-03-23
c: Try loading OpenGL through EGL if GLX isn't available

ziyao233:
On a UNIX system without X11 libraries, it's impossible to have libGL.so and glXGetProcAddressARB(). Currently GLAD would fail to load GL symbols in this case.

Try also loading libEGL.so/libEGL.so.1 in OpenGL loader and fallback to eglGetProcAddress() to query GL symbols if glXGetProcAddressARB() isn't available.

Since EGL version 1.5 released in 2014, it's required for eglGetProcAddress() to be able to query all client API functions instead of only extensions, which is quite early for a Wayland-only display stack, thus no function check or another fallback logic is added.

### DIFF
--- a/glad/generator/c/templates/loader/gl.c
+++ b/glad/generator/c/templates/loader/gl.c
@@ -44,7 +44,9 @@ static void* glad_gl_dlopen_handle({{ template_utils.context_arg(def='void') }})
         "libGL-1.so",
   #endif
         "libGL.so.1",
-        "libGL.so"
+        "libGL.so",
+        "libEGL.so.1",
+        "libEGL.so"
     };
 #endif
 
@@ -67,6 +69,9 @@ static struct _glad_gl_userptr glad_gl_build_userptr(void *handle) {
 #else
     userptr.gl_get_proc_address_ptr =
         (GLADglprocaddrfunc) glad_dlsym_handle(handle, "glXGetProcAddressARB");
+    if (!userptr.gl_get_proc_address_ptr)
+        userptr.gl_get_proc_address_ptr =
+            (GLADglprocaddrfunc) glad_dlsym_handle(handle, "eglGetProcAddress");
 #endif
 
     return userptr;


Dav1dde:
I wonder though if this should be an approach similar to what the GLES loader does, depend on `glad/egl.h` to provide `eglGetProcAddress` instead of trying to load it from the `.so`.

Relying on `glad/egl.h` seems like the more resilient approach, but this now means we'd have to probe whether it is available.

ziyao233:
> I wonder though if this should be an approach similar to what the GLES loader does, depend on `glad/egl.h` to provide `eglGetProcAddress` instead of trying to load it from the `.so`.

I don't have a strong opinion on this. Existing code in `gl.c` loads `*GetProcAddress()` symbols by directly poking around `libGL.so` or `opengl32.dll`, so for me it feels natural to do the similar things.

> Relying on `glad/egl.h` seems like the more resilient approach, but this now means we'd have to probe whether it is available.

I'm willing to do so if you prefer. In this case, should we also refactor `gl.c` to depend on `glad/{wgl,glx}.h` before adding EGL support into it? This would keep the style consistent and reduce some code duplication, I think.

Dav1dde:
I think the difference for `wgl` is that `wglGetProcAddress` was always able to load GL symbols, while I think `egl` only worked for GLES. I am not sure if that was changed or if that is still the case.

Do you happen you know where in the egl spec it says it can load OpenGL symbols and with what version and extensions it requires?

> I'm willing to do so if you prefer [..]

I think your approach is fine to start out with, as long as the spec confirms this should work. I am just worried about introducing cases where it just silently breaks.

ziyao233:
> I think the difference for `wgl` is that `wglGetProcAddress` was always able to load GL symbols, while I think `egl` only worked for GLES. I am not sure if that was changed or if that is still the case.
> 
> Do you happen you know where in the egl spec it says it can load OpenGL symbols and with what version and extensions it requires?

Since EGL 1.4 it clearly claims support for OpenGL, quoting EGL specfication 1.5 Appendix E.1 "Updates to EGL 1.4",

> Changes in the revision approved on February 13, 2013:
> • Clarify support for OpenGL as well as OpenGL ES in sections 1, 2.2, 2.2,
> 2.2.2, 2.3, 2.4, 2.6, and 5 (Bug 9864).

Also since EGL 1.4, as mentioned in the commit message, `eglGetProcAddress()` is able to query EGL and client API functions, instead of only ones defined by extensions. 

> – Defining eglGetProcAddress to support querying of all EGL
> and client API functions, not just extensions (EGL_KHR_get_-
> all_proc_addresses and EGL_KHR_client_get_all_proc_-
> addresses).

and in a Wayland-only UNIX environment, we don't have other choices than EGL. As mentioned in the commit message, GLX entrypoints (of course) require X11 libraries to compile. We do have libglvnd providing wrapper `libOpenGL.so` which isn't dependent on X11, but it only ships OpenGL 4.5 Core Profile symbols[1].  

> > I'm willing to do so if you prefer [..]
> 
> I think your approach is fine to start out with, as long as the spec confirms this should work. I am just worried about introducing cases where it just silently breaks.

Thanks. I don't think changes in this pull request will make any real-world use cases worse than current.

The newly added EGL path only gets executed when `glXGetProcAddressARB()` isn't present, which means GLX isn't available. In this case, we currently use `dlsym()` to load GL symbols, and this could only work if the application is somehow linked to a library providing all GL extensions it requires.

[1]: https://gitlab.freedesktop.org/glvnd/libglvnd/-/work_items/63

Dav1dde:
Thanks for collecting all the infos!

PR #538, opened 2026-04-10
Add --extension-file option to merge additional XML files into the specification

chrisd-work:
Developers working on Vulkan features not yet merged into vk.xml had no way to test against the spec without modifying it directly.

This adds an --extension-file option that accepts a path (or comma-separated list of paths) to additional XML files, which are merged into vk.xml at runtime — allowing users to prototype and test new extensions without touching the upstream spec.

Usage:
glad --extension-file VK_test_extension.xml ...
glad --extension-file VK_ext_a.xml,VK_ext_b.xml ...

### DIFF
--- a/glad/parse.py
+++ b/glad/parse.py
@@ -279,6 +279,24 @@ class Specification(object):
         except TypeError:
             return cls(xml_parse(path_or_file_like))
 
+    def merge_from_file(self, path):
+        """
+        Merge additional XML elements from an extension file into this specification.
+        Appends children from <types>, <commands>, <extensions>, and <platforms>
+        sections of the extension file into the corresponding sections of this spec.
+        """
+        ext_root = xml_parse(path)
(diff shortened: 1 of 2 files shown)


glad/parse.py
```diff
@@ -279,6 +279,24 @@ def from_file(cls, path_or_file_like, opener=None):
...
             return cls(xml_parse(path_or_file_like))
 
+    def merge_from_file(self, path):
+        """
+        Merge additional XML elements from an extension file into this specification.
+        Appends children from <types>, <commands>, <extensions>, and <platforms>
+        sections of the extension file into the corresponding sections of this spec.
+        """
+        ext_root = xml_parse(path)
+        for section_tag in ('types', 'commands', 'extensions', 'platforms'):
```
Dav1dde, line 289: Out of curiosity, why can't the entire document be merged unconditionally?

Dav1dde (approved):

PR #540, opened 2026-06-18
ci: fix build and bump action versions

scribam:
- Bump actions/checkout and actions/setup-python versions to be compatible with Node 24 (https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/)
- Skip pip upgrade and cache pip dependencies
- Install a newer wine version to fix the build (from https://gitlab.winehq.org/wine/wine/-/wikis/Debian-Ubuntu)


Dav1dde:
Awesome, thanks a lot!

